PR #64, opened 2017-10-14
Refactored Stylesheets

Trevor-Welch:
style: Added custom mixin to load themes cleanly.

Added a Theme folder for all theme partials to be broken down and
imported as needed, as well as, wrote a custom mixin 'load-theme( )'
which cleans up the code a bit and prevents repetition.  I also changed
the font-weight of the paragraph text to 300 for better readability as
the current weight was very heavy.


src/theme/teamcraft-theme.scss
```diff
@@ -0,0 +1,88 @@
+//Importing this file allows the use of the $teamcraft-theme variable
+//You can use this variable in every way as $light-theme or $dark-theme
+//Updated 10/13/2017 by Delaxiox
+
+//   Custom Function - Creates map of a single color hex
+@function hex-to-mat-color($hex){
```
Supamiu: I'm not a big fan of this, the different palette shades are actually used in some components, overriding it for a single-color palette seems kinda strange to me (if I understand well)

src/styles.scss
```diff
@@ -1,16 +1,46 @@
...
+// Version Text
+p.version.light-text{
+  top: 0px!important;
+  z-index: 99;
+}
+
+// Make header bar shorter
+md-toolbar.mat-toolbar.mat-primary{
+  min-height: 48px!important;
+  box-shadow: 0 3px 5px -1px rgba(0,0,0,.2), 0 6px 10px 0 rgba(0,0,0,.14), 0 1px 18px 0 rgba(0,0,0,.12);
```
Supamiu: This can be replaced by a `.mat-elevation-zX` with X in [1,24]. See https://stackoverflow.com/questions/41534396/how-to-define-an-elevation-in-angular-material-2

Supamiu:
I'm closing it for now, as you're planning to add more changes, opening a new PR once the changes have been made will be better.

PR #93, opened 2017-11-01
feat: added a button to reset filters

ravenjolly:


src/app/component/list-details/list-details.component.html
```diff
@@ -109,11 +109,19 @@
...
+                        <button md-raised-button
+                                [disabled]="(getUser() | async)?.isAnonymous"
+                                (click)="adaptFilters()"
+                                color="accent">Adapt to character
+                        </button>
+
+                        <button md-raised-button
+                                [disabled]="(getUser() | async)?.isAnonymous"
+                                (click)="resetFilters()"
+                                color="accent">Reset Filters
```
Supamiu: I think that you should add translation here, and add it to the "Adapt to character" button to at the same time ;)

PR #98, opened 2017-11-14
feat: Persistent sidebar on larger devices

ravenjolly:
I've targeted it to persist for tablets(768px) or larger devices that have the screen space to host the bar and the content.


src/app/app.component.ts
```diff
@@ -51,6 +54,17 @@ export class AppComponent implements OnInit {
                 private userService: UserService,
                 private snack: MatSnackBar) {
 
+
+        //check for width
+        if (window.screen.width >= 768) {
```
Supamiu: You should use ObservableMedia instead, see https://github.com/Supamiu/ffxiv-teamcraft/blob/refactor/src/app/modules/item/item/item.component.ts#L285. Documentation for ObservableMedia: https://github.com/angular/flex-layout/wiki/Responsive-API

src/app/app.component.ts
```diff
@@ -196,4 +210,18 @@ export class AppComponent implements OnInit {
         localStorage.setItem('locale', lang);
         this.translate.use(lang);
     }
+
+    @HostListener('window:resize', ['$event'])
```
Supamiu: Can be removed if you use ObservableMedia.

PR #160, opened 2017-12-31
Style: Correct currency name display.

jco2641:
The plural form of 'gil' is 'gil'


Supamiu:
Hi, thank you for this first contribution 👍 

Your changes will be included in 3.1.

PR #163, opened 2018-01-10
feat: 10k lists giveway popup

Supamiu:


src/assets/i18n/de.json
```diff
@@ -8,6 +8,11 @@
         "Patreon_supporters": "Patreon-Unterstützer",
         "Contributors": "Mitwirkende"
     },
+    "GIVEWAY": {
```
Supamiu: @Nivrelin Can you please translate the content of this? 😃

Issue #169, opened 2018-01-15
chore: 3.2 translation


Supamiu:
@Nivrelin I'll reference this issue in commits including things to translate for 3.2, this way you'll see every translations for 3.2 at the same place.

Supamiu:
https://github.com/Supamiu/ffxiv-teamcraft/commit/1666c358fad4421ef1cacd9ff8939d7a1d5fa193

Supamiu:
https://github.com/Supamiu/ffxiv-teamcraft/commit/07da2410d1a50f3cc918b53923a9763ef8bc842b

This should be the last one, once we have everything translated it should be ok for 3.2

PR #175, opened 2018-01-19
feat: custom layout for list details

### LINKED ISSUE #168: feat: custom layouts for list details
The intent is to provide a way for users to create their own filters for data panels, with a default layout for everyone, that can be modified easily.

Supamiu:
closes #168
#169


Supamiu:
@Nivrelin please check changes for de.json, I added one common string at the root of the JSON, be careful not to miss it 😄 
Same for @AltanRapsied but for fr.json file.

PR #308, opened 2018-04-03
fix: isearch value quotes

i5nicca:
I suppose /isearch value should be quoted. It behaves wrong for items that name contains more than one word.


Supamiu:
Hey, thank you for your PR :) I first wanted to fix this using `\"` but using `&quot;` like you did feels better imo. will be live very soon

PR #370, opened 2018-05-25
fix: fixes missing linebreaks when copying macro from firefox

Creatan:
Adding a linebreak element for each macro action fixes an firefox issue where linebreaks were not present when copying macros.


Supamiu:
Thanks for your contribution, this will be added in 4.0.8, happening either this weekend or Monday.

PR #379, opened 2018-05-28
fix: missing linebreaks in cross-class setup macro

Creatan:
Cross-class skill setup macro was also missing line-breaks when copying from Firefox.


Supamiu:
Oh, I forgot that :O thanks for the fix 😄

PR #413, opened 2018-06-20
chore: add SaintCoinach as submodule

mihawk90:
Adding SaintCoinach as a submodule to reduce work when updates are coming in, no manual copying necessary anymore.

XIVExtractor had to be moved to parent directory to be able to track it in TC repository while maintaining a clean copy of SC.


mihawk90:
Eeehhh something bad happened apparently... let me re-do this PR xD

PR #414, opened 2018-06-20
chore: add SaintCoinach as submodule

mihawk90:
Adding SaintCoinach as a submodule to reduce work when updates are coming in, no manual copying necessary anymore.

XIVExtractor had to be moved to parent directory to be able to track it in TC repository while maintaining a clean copy of SC.

Cherry-Picked from #413


Supamiu:
We need to find a fix for travis before merging this, mainly because travis build is now relevant, it tests that the build is going fine, having it fail on each PR is not something I want at all :/ going to investigate it, if you find a solution, please commit it 😄

mihawk90:
It seems to be related to access rights, not having permission to read. I'm not sure why though because it's a public repo and the regular github login...

PR #421, opened 2018-06-23
feat: apply stats to all crafts button

### LINKED ISSUE #418: feat: apply stats to all crafts button
## I'm submitting a...
* (_) Bug report *- please add **bug:** before the title*
* (x) Feature request *- please add **feat:** before the title*
* (_) Support request *- please come on [our discord](https://discord.gg/r6qxt6P) instead*

## What feature are you proposing

The ability to apply stats for all crafters at once, from the crafter stats dialog box.

## What is the use case for this feature?

As gear is often shared between crafters when talking about endgame gearing, being able to save given stats for all crafters at once can be a really nice thing to have to avoid updating all crafters one by one, resulting in 8 dialog boxes to open and a lot of user interactions to paste the same thing.

mattantonelli:
Adds a "save for all" button, allowing users to quickly save the stats for one gearset across all gearsets. Resolves #418

### DIFF
--- a/src/app/pages/profile/profile/profile.component.ts
+++ b/src/app/pages/profile/profile/profile.component.ts
@@ -26,7 +26,7 @@ import {combineLatest, of} from 'rxjs';
 })
 export class ProfileComponent extends PageComponent {
 
-    static craftingJobs = [
+    private craftingJobs: any[] = [
         {abbr: 'CRP', name: 'carpenter'},
         {abbr: 'BSM', name: 'blacksmith'},
         {abbr: 'ARM', name: 'armorer'},
@@ -69,7 +69,7 @@ export class ProfileComponent extends PageComponent {
                             });
                         }),
                         map(sets => sets.map(set => {
-                                const job = ProfileComponent.craftingJobs[set.jobId - 8];
+                                const job = this.craftingJobs[set.jobId - 8];
                                 if (job !== undefined) {
                                     set.abbr = job.abbr;
                                     set.name = job.name;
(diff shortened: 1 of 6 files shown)


mihawk90:
Should probably be merged to 4.3.0 branch instead of master directly.

src/app/pages/profile/profile/profile.component.ts
```diff
@@ -111,7 +111,7 @@ export class ProfileComponent extends PageComponent {
     }
 
     public openStatsPopup(set: GearSet): void {
-        this.dialog.open(StatsEditPopupComponent, {data: set});
+        this.dialog.open(StatsEditPopupComponent, {data: {set: set, jobs: this.craftingJobs.slice(0, 8)}});
```
Supamiu: Why do you slice `this.craftingJobs` here? is it to pass a copy?

src/app/pages/profile/profile/profile.component.ts
```diff
@@ -26,7 +26,7 @@ import {combineLatest, of} from 'rxjs';
 })
 export class ProfileComponent extends PageComponent {
 
-    static craftingJobs = [
+    private craftingJobs: any[] = [
```
Supamiu: iirc, this was set to static because it's used somewhere else, have you checked that is wasn't?

Supamiu: Just checked, it's not used anywhere else. nvm

src/app/pages/profile/profile/profile.component.ts
mattantonelli: `craftingJobs` is actually a misnomer because the last 3 objects are DoL jobs. The slice here ensures we are only passing in DoH jobs.

PR #526, opened 2018-07-27
refactor: recipe search striping

mattantonelli:
Striping odd rows instead of even rows. Lowered the opacity of the striping to provide a less harsh contrast, especially for light themes.


Supamiu:
I changed base to 4.4 as I'd like to ship that in 4.4 update rather than in a 4.3.10

PR #553, opened 2018-08-07
fix(simulator): Fix actions being placed into wrong spots

### LINKED ISSUE #552: bug(simulator): shifting actions in a rotation to the right puts it in the wrong spot
**Describe the bug**
Moving actions in a rotation puts them 1 spot to the right of where they should be.

**To Reproduce**
1) Create any rotation of at least 3 actions
2) Move first action between 2nd and 3rd
3) Action will appear after the 3rd one instead

**Expected behavior**
Actions should go where the mouse is pointing

**Additional context**
Since this only appears when moving to the right, it's likely caused by the indexing being changed. E.g. it's recorded that the action should go before the action with index 6, but that action is shifted to the left to fill the gap, so now its index is 5 and the action is inserted after it.

Cypher121:
Change drag-and-drop logic to manually move part of the array between the old and new position.
Should also be a bit faster than 2 splices.

Closes #552

### DIFF
--- a/src/app/pages/simulator/components/simulator/simulator.component.html
+++ b/src/app/pages/simulator/components/simulator/simulator.component.html
@@ -436,7 +436,7 @@
                 </div>
                 <div class="action-drag-zone" droppable
                      dragOverClass="drag-over"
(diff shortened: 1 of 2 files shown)


Supamiu:
It seems like circleCI got an issue with the webhook for this PR, I'll accept it and will test on beta.

PR #610, opened 2018-10-03
fix: consumable styles

mattantonelli:
Fixes incorrect flexing on consumable styles where the widths were incorrect and the placeholder text was not being displayed.

**Before:**
image

**After:**
image


Supamiu:
If you're using vscode for this branch, it has a nice extension for tslint integration, as linting is required for a build to happen. https://marketplace.visualstudio.com/items?itemName=eg2.tslint

PR #611, opened 2018-10-03
feat: convert custom rotation to normal rotation

### LINKED ISSUE #609: feat:Turn Custom Rotation in to Normal Rotation
**Is your feature request related to a problem? Please describe.**
With the creation of custom rotations.  we can save a rotation with custom stats. but there is no easy way to change a Custom rotation to a Regular (item based rotation)

**Describe the solution you'd like**
A way to turn a rotation that was saved with the custom feature in to a one based off a item. (real or custom) might be a good transition for 5.0
like say you have a 380 rotation. from before the item lists or such was updated

**Describe alternatives you've considered**
Well there is the long path of searching through the item list and making a new rotation based of the old one but that pretty lengthy and some what annoying.

**Additional context**
Additional items would be to make  a companion code or such for when 5.0 comes out to make it accept  the new (custom) item system.  

and a bit from Rae:
1. Click button to save as normal rotation
2. Select item from search
3. ???
4. Profit

mattantonelli:
Adds a button to the Simulator page that allows users to convert a saved
custom rotation into a normal rotation. This is done by allowing the
user to search for a recipe to use for the rotation, and then saving
that item/recipe with the rotation so it is opened like a normal one.

Resolves #609

### DIFF
--- a/src/app/pages/simulator/components/simulator/simulator.component.ts
+++ b/src/app/pages/simulator/components/simulator/simulator.component.ts
@@ -559,7 +559,7 @@ export class SimulatorComponent implements OnInit, OnDestroy {
                         authorId: this.authorId,
                         consumables: { food: this._selectedFood, medicine: this._selectedMedicine },
                         freeCompanyActions: this._selectedFreeCompanyActions,
-                        folder: ''
+                        folder: this.rotation === undefined ? '' : this.rotation.folder
                     });
                 })
             });
--- a/src/app/pages/simulator/components/simulator/simulator.component.ts
+++ b/src/app/pages/simulator/components/simulator/simulator.component.ts
@@ -543,6 +543,29 @@ export class SimulatorComponent implements OnInit, OnDestroy {
         }
     }
 
+    convertRotation(): void {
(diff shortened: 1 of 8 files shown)


src/app/pages/simulator/components/simulator/simulator.component.ts
```diff
@@ -543,6 +543,29 @@ export class SimulatorComponent implements OnInit, OnDestroy {
...
+                        $key: this.rotation.$key,
+                        name: this.rotation.name,
+                        rotation: this.serializedRotation,
+                        recipe: item.getCraft(result.recipeId),
+                        defaultItemId: result.itemId,
+                        defaultRecipeId: result.recipeId,
+                        authorId: this.authorId,
+                        consumables: { food: this._selectedFood, medicine: this._selectedMedicine },
+                        freeCompanyActions: this._selectedFreeCompanyActions,
+                        folder: ''
```
Supamiu: This should be the folder used for current rotation, not an empty one (which will pop the rotation out of its current folder upon conversion)

PR #750, opened 2019-02-08
Fest/add chinese simple item data

gsmsh:
I add the Chinese simple item data to "data.json", but it seems that the order of the original data is messed up, because the code language I used. Does this matter very much？ In addition, I never used Angular.js before. If I edited  something wrong, please let me know.


apps/client/src/app/pages/search/search/search.component.ts
```diff
@@ -77,7 +77,7 @@ export class SearchComponent implements OnInit {
               private rotationPicker: RotationPickerService, private htmlTools: HtmlToolsService,
               private message: NzMessageService, private translate: TranslateService) {
     this.uiCategories$ = this.xivapi.getList(XivapiEndpoint.ItemUICategory, {
-      columns: ['ID', 'Name_de', 'Name_en', 'Name_fr', 'Name_ja'],
+      columns: ['ID', 'Name_de', 'Name_en', 'Name_fr', 'Name_ja', 'Name_zh'],
```
Supamiu: Unfortunately, this won't be working because xivapi doesn't provide `Name_zh`, that being said, I should convert this to a basic usage of the itemName pipe inside the template. I'll take care of this part.

Supamiu:
Hello, thank you for your contribution ! 

That's an impressive new PR, really good stuff 👍 

May I ask you how you got the chinese names for the items? I would like to find a way to automate this process for future updates, as the items.json file is generated, the zh entries might be removed on next update :/

Supamiu:
While your PR is nice and you did a good job, this won't enable chinese inside search system, reason being that the api behind that does search for us doesn't have zh names, see https://www.garlandtools.org/api/search.php?type=item&lang=zh&text=%E9%AA%91%E4%BE%A0%E5%92%8F%E5%92%92%E8%85%95%E7%8E%AF

gsmsh:
> Hello, thank you for your contribution !
> 
> That's an impressive new PR, really good stuff 👍
> 
> May I ask you how you got the chinese names for the items? I would like to find a way to automate this process for future updates, as the items.json file is generated, the zh entries might be removed on next update :/

Thank you for your reply! It is acceptable for me to use English for searching and show the items in Chinese.In that way, I can make a crafting list and share it to my friends who cannot use Ehglish. But, I will also look for some ways for searing in Chinese.

As for the item names, because the SaintCoinach can't work well with Chinese client , I use a fork of SaintCoinach to extract these names and save them in a .csv file. Then, I use golang to add each one to items.json,. This is the fork of SaintCoinach for chinese client. Hope it can be helpful.
[https://github.com/thewakingsands/SaintCoinach](url)

Supamiu:
Maybe it would be easier to just provide a zh-items.json file so your work isn't overwritten on next update, and you can PR the update whenever you want. 

The logic can then be combined inside the LocalizedDataService, to get zh name after the base i18n name and combine them into a complete I18nName entry

gsmsh:
That sounds great. I will close this PR for now, and submit a new one later. Thank you for your advise :)

PR #766, opened 2019-02-17
feat: support Chinese item names

gsmsh:
I add a 'zh-items.json' file which includes only zh items' names. When a item row is requested, the LocalizedDataService would combine this with 'items.json' and return the result.


apps/client/src/app/model/garland-tools/item.ts
```diff
@@ -12,6 +12,7 @@ export class Item implements I18nData {
   en: I18nDataRow;
   de: I18nDataRow;
   ja: I18nDataRow;
+  zh: I18nDataRow;
```
Supamiu: garland tools doesn't provide zh item names, same for ko, which is why ko isn't in this interface ^^

Supamiu:
Really nice PR, thanks a lot 😄 

Clean changes, you understood my code and applied nice updates properly :)

apps/client/src/app/model/garland-tools/item.ts
gsmsh: I must forget this. Thanks and I have fixed this.

PR #785, opened 2019-02-26
 feat: support Korean item names

Ra-Workspace:
I refered PR #766 , It was so hard for me!


Supamiu:
Very good first PR, thank you for your work :)

Issue #1071, opened 2019-08-06
feat: amount of list text export not consider the done amount


transparentmask:
*please prefix the issue with "feat:", example: "feat: commission board system"*

**Is your feature request related to a problem? Please describe.**
When I do text export in a list, the amount of items is total amount. But in fact, in my list, there are some items mark 'done' or 'partially done'. So, I think maybe remaining amount is better.

For example, in my list needs total 30 "water crystals", and some items already done. So, I just need 20 "water crystals" now. At this time, after I click "text export" button in list, the result in my clipboard still show "30x water crystals", but I think it should tell me "20x water crystals" for better.

**Describe the solution you'd like**
In the function getListTextExport of apps/client/src/app/pages/list-details/list-details/list-details.component.ts, it's only using the ${row.amount}.
For this issue, may be using ${row.amount-row.done} instead of ${row.amount}.

**Additional context**
This may be a very personal feature, but I think that at least two export methods can be made optional.

Thank you.

PR #1072, opened 2019-08-06
feat: amount of list text export not consider the done amount

transparentmask:
Changes for my issue https://github.com/ffxiv-teamcraft/ffxiv-teamcraft/issues/1071


Supamiu:
Hello, thank you for your PR :)

Did you test these changes? I don't think `row.amount-row` exists at all and you'll probably end up with `null` inside your export string.

transparentmask:
I tested some situations, but it may not cover all of them.
In my tests, I see the default value of row.done is 0.
And I noticed, in function upgradeList of apps/client/src/app/modules/list/list-manager.service.ts, 
listRow.used = row.item.used || 0;

May be I should do like this. I will close this and do another PR.

PR #1249, opened 2019-11-26
chore: Opcode updates node-machina-ffxiv

karashiiro:
This has the autodetected packet opcodes, the rest are coming soon^TM


karashiiro:
All but initZone, persistentEffect, and event packets

PR #1301, opened 2019-12-24
Added a "Copy to clipboard" to the crystals panel within lists

TwilightVanish:


apps/client/src/app/pages/list-details/list-crystals-panel/list-crystals-panel.component.ts
```diff
@@ -12,8 +16,22 @@ export class ListCrystalsPanelComponent {
...
 
+  constructor(private i18nTools: I18nToolsService, private l12n: LocalizedDataService, private message: NzMessageService, private translate: TranslateService) {};
+
   trackByItem(index: number, item: ListRow): number {
     return item.id;
   }
 
+  public getTextExport(): string {
+    let exportString = "Crystals:\n";
+    for (const crystal in this.crystals) {
```
Supamiu: Please use `Object.keys` instead, making it safer and will allow it to pass the linter (linter asks for `hasOwnProperty` check, which is here to make it safe).


See https://github.com/ffxiv-teamcraft/ffxiv-teamcraft/blob/staging/apps/client/src/app/modules/list/list-details-panel/list-details-panel.component.ts#L370-L382 for a full example.

apps/client/src/app/pages/list-details/list-crystals-panel/list-crystals-panel.component.ts
```diff
@@ -12,8 +16,22 @@ export class ListCrystalsPanelComponent {
...
   crystals: ListRow[] = [];
 
+  constructor(private i18nTools: I18nToolsService, private l12n: LocalizedDataService, private message: NzMessageService, private translate: TranslateService) {};
+
   trackByItem(index: number, item: ListRow): number {
     return item.id;
   }
 
+  public getTextExport(): string {
+    let exportString = "Crystals:\n";
```
Supamiu: This needs to be translated, use `${this.translate.instant('Crystals')}` instead.

Supamiu:
Thank you for your PR :) minor changes are needed but other than that, it's pretty good !

PR #1314, opened 2019-12-31
FIX: Don't suggest IQ when TE is used

Strati:
Inner Quiet should not be suggested when Trained Eye is used as the first step to max quality.


ayyaruq:
Elegantly done, almost like I wrote it myself.

PR #1326, opened 2020-01-18
Update Chinese data & optimize i18n filling for ZH/KR

zhyupe:


apps/client/src/app/core/data/localized-data.service.ts
```diff
@@ -292,6 +227,47 @@ export class LocalizedDataService {
     return array[id];
   }
 
+  private getRowWithExtendedLanguage(key: keyof LazyData, id: number | string): I18nName {
+    let row = this.getRow<I18nName>(this.lazyData.data[key], id);
```
Supamiu: Because row is never reassigned, you should use `const` instead of `let` (this breaks linter in CI).
```suggestion
    const row = this.getRow<I18nName>(this.lazyData.data[key], id);
```

apps/client/src/app/core/data/localized-data.service.ts
```diff
@@ -292,6 +227,47 @@ export class LocalizedDataService {
...
+
+  private guessExtendedLanguageKeys(key: keyof LazyData) {
+    return {
+      zhKey: this.guessExtendedLanguageKey('zh', key),
+      koKey: this.guessExtendedLanguageKey('ko', key)
+    };
+  }
+
+  private guessExtendedLanguageKey(language: 'zh' | 'ko', key: keyof LazyData): keyof LazyData {
+    let guessKey = `${language}${key.charAt(0).toUpperCase()}${key.substr(1)}`
```
Supamiu: Because guessKey is never reassigned, you should use `const` instead of `let` (this breaks linter in CI).
```suggestion
    const guessKey = `${language}${key.charAt(0).toUpperCase()}${key.substr(1)}`
```

Supamiu:
👋 Thank you rof this amazing PR, you really did a good job and overall, only small lint fixes are needed.

Please feel free to reach me on Discord if you want to do more HQ PRs like this one and need some input regarding the code.

PR #1327, opened 2020-01-18
better translation for KR

Jaehyuk-Lee:
previous: star = 20% more exp
after: star = 20% more reward


Supamiu:
Thank you for your contribution :)

For future translation fixes, please use crowdin instead, as it creates the PR for you and makes sure everything is in a sync state: https://crowdin.com/project/ffxiv-teamcraft

PR #1354, opened 2020-02-14
DB localization enhancement

zhyupe:
1. add DB used by geatsets
2. add ability to convert xivapi result to `I18nName`, so names of extended languages can be displayed with xipapi result now
3. add Chinese & Korean to i18n name list (placed in the right of DB page), and make it a component to reduce duplicate codes
image
4. add Chinese & Korean to patch, item descriptions, placenames, etc
5. optimizations for levequest page


Supamiu:
That's an amazing PR, thank you for all this, the work is really well done and you refactored stuff I did too fast without thinking about it.

PR #1355, opened 2020-02-15
feat(simulator): update for 5.2 preliminary patch notes

ayyaruq:
Placeholder icons - Flawless for Groundwork, Innovative Touch for Veneration.


ayyaruq:
Build is broken because a git reference can't be handled by webpack when it needs compilation. I've done a build and `npm link` locally to verify this and it looks good, but prob best to do a proper release of the sim lib on npm and then update the `package.json` for that, which I guess means waiting for xivapi to get the new skill IDs and icons.

We may need new rotation tips but we have nothing for Innovation or Preparatory, so unlikely to need anything for Veneration and Groundwork. Maybe a future feature.

Supamiu (approved):

PR #1372, opened 2020-02-21
Added "Consolidate Stacks" Inventory Optimizer

rcdailey:
A new inventory optimizer that will suggest high-quality (HQ) stacks of
items that can have their quality lowered so that it can be joined with
an existing normal-quality (NQ) stack of that same type of item. This is
a very reliable way to free up inventory slots.

Because users may not want all stacks of items suggested, filtering is
provided by expansion. For example, if a user wants all items in
Heavensward and below suggested (Heavensward + A Realm Reborn), they
will select "Heavensward" from the filter select box. Items with HQ and
NQ stacks from Stormblood and Shadowbringers will *not* be selected in
this case because they are filtered out. You can also clear the filter
box (so it says "Pick Expansion") which completely disables filtering,
showing all available items in your inventory to which the optimization
applies.

In general, if item database information is incomplete, an item does not
get filtered even if it may have been otherwise. Examples:

* The specified item does not have an assigned patch.
* A patch to which an item is assigned is not defined.

Other changes:

* Debugging support for VS Code

* Don't ignore dupes at same slot in diff inventory

  The duplicate optimizer no longer ignores items in the same slot but in
different inventories. The condition used to check only if the slot ID
was the same. Now in addition to that, the inventory ID must also match.

* Optimizer panels no longer hide when empty

  For optimizer panels with interactive input fields on them, if they are
hidden due to the filters being too strict, that makes it so that the
user will not be able to alter the values.

  To avoid this, the panels are no longer hidden if they are empty. The
user can refer to the count presented on the panel text itself to
determine if they should ignore it or not (e.g. zero items).

* Add gitattributes

### DIFF
--- a/apps/client/src/app/pages/inventory-optimizer/inventory-optimizer/inventory-optimizer.component.ts
+++ b/apps/client/src/app/pages/inventory-optimizer/inventory-optimizer/inventory-optimizer.component.ts
@@ -13,6 +13,7 @@ import { NzMessageService } from 'ng-zorro-antd';
 import { TranslateService } from '@ngx-translate/core';
 import { HasTooFew } from '../optimizations/has-too-few';
 import { ListRow } from '../../../modules/list/model/list-row';
+import { ConsolidateStacks } from '../optimizations/consolidate-stacks';
 
 @Component({
   selector: 'app-inventory-optimizer',
@@ -119,12 +120,37 @@ export class InventoryOptimizerComponent {
   }
 
   constructor(private inventoryFacade: InventoryFacade,
-              @Inject(INVENTORY_OPTIMIZER) private optimizers: InventoryOptimizer[],
-              private lazyData: LazyDataService, private message: NzMessageService, private translate: TranslateService) {
+    @Inject(INVENTORY_OPTIMIZER) private optimizers: InventoryOptimizer[],
+    private lazyData: LazyDataService, private message: NzMessageService, private translate: TranslateService) {
   }
 
   private getContainerName(item: InventoryItem): string {
     return item.retainerName || this.inventoryFacade.getContainerName(item.containerId);
+  }
+
+  public getExpansions(): string[] {
+    const expansions: any[] = this.lazyData.patches.map(p => {
+      return { name: p.ExName, version: p.ExVersion }
+    });
+
+    return uniqBy(expansions, 'name');
+  }
+
+  public get selectedExpansion(): number {
+    const selection = localStorage.getItem(ConsolidateStacks.SELECTION_KEY);
+    return selection ? +selection : null;
+  }
+
+  public set selectedExpansion(selection: number) {
+    if (selection !== null) {
+      localStorage.setItem(ConsolidateStacks.SELECTION_KEY, selection.toString());
+    }
+    else {
+      localStorage.removeItem(ConsolidateStacks.SELECTION_KEY);
+    }
+
+    this.loading = true;
+    this.resultsReloader$.next(null);
   }
 
   nameCopied(key: string, args?: any): void {
--- a/apps/client/src/app/pages/inventory-optimizer/inventory-optimizer/inventory-optimizer.component.ts
+++ b/apps/client/src/app/pages/inventory-optimizer/inventory-optimizer/inventory-optimizer.component.ts
@@ -57,8 +57,7 @@ export class InventoryOptimizerComponent {
                     .value(),
                   totalLength: uniqBy(entries, 'item.itemId').length
                 };
-              })
-              .filter(res => res.entries.length > 0);
+              });
           })
         )),
         tap(() => this.loading = false)
@@ -95,7 +94,7 @@ export class InventoryOptimizerComponent {
             });
             opt.totalLength = uniq(total).length;
             return opt;
-          }).filter(opt => opt.totalLength > 0);
+          });
         })
       );
     })
--- a/apps/client/src/app/pages/inventory-optimizer/optimizations/consolidate-stacks.ts
+++ b/apps/client/src/app/pages/inventory-optimizer/optimizations/consolidate-stacks.ts
@@ -1,0 +1,79 @@
+import { InventoryOptimizer } from './inventory-optimizer';
+import { InventoryItem } from '../../../model/user/inventory/inventory-item';
+import { UserInventory } from '../../../model/user/inventory/user-inventory';
+import { ListRow } from '../../../modules/list/model/list-row';
+import { TranslateService } from '@ngx-translate/core';
+import { InventoryFacade } from '../../../modules/inventory/+state/inventory.facade';
+import { LazyDataService } from '../../../core/data/lazy-data.service';
+import { Memoized } from '../../../core/decorators/memoized';
+
+export class ConsolidateStacks extends InventoryOptimizer {
+
+  static SELECTION_KEY = 'optimizer:consolidate-stacks:selection';
+
+  constructor(private translate: TranslateService, private inventoryFacade: InventoryFacade, private lazyData: LazyDataService) {
+    super();
+  }
+
+  @Memoized()
+  getPatch(id: number): any {
+    return this.lazyData.patches.find(p => p.ID === id);
+  }
+
+  itemInExpansion(itemId: number, expansion: number): boolean {
+    // If the specified item does not have a patch associated with it, we do not filter it.
+    const itemPatch = this.lazyData.data.itemPatch[itemId];
+    if (itemPatch === null) {
+      return true;
+    }
+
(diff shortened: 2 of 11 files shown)


apps/client/src/app/pages/inventory-optimizer/inventory-optimizer/inventory-optimizer.component.ts
```diff
@@ -120,14 +120,41 @@ export class InventoryOptimizerComponent {
...
   }
 
+  public getExpansions(): string[] {
+    let expansions: any[] = this.lazyData.patches.map(p => {
+      return { name: p.ExName, version: p.ExVersion }
+    });
+
+    return expansions.filter((value, index, array) => {
+      return index === array.findIndex(e => e.name == value.name);
+    });
```
Supamiu: Could probably be changed for a `uniqBy(expansions, 'name');` call, using `uniqBy` from lodash.

apps/client/src/app/pages/inventory-optimizer/inventory-optimizer/inventory-optimizer.component.ts
```diff
@@ -120,14 +120,41 @@ export class InventoryOptimizerComponent {
...
+  }
+
+  public get selectedExpansion(): number {
+    let selection = localStorage.getItem(ConsolidateStacks.SELECTION_KEY);
```
Supamiu:
```suggestion
    const selection = localStorage.getItem(ConsolidateStacks.SELECTION_KEY);
```

apps/client/src/app/pages/inventory-optimizer/inventory-optimizer/inventory-optimizer.component.ts
```diff
@@ -120,14 +120,41 @@ export class InventoryOptimizerComponent {
...
+    @Inject(INVENTORY_OPTIMIZER) private optimizers: InventoryOptimizer[],
+    private lazyData: LazyDataService, private message: NzMessageService, private translate: TranslateService) {
   }
 
   private getContainerName(item: InventoryItem): string {
     return item.retainerName || this.inventoryFacade.getContainerName(item.containerId);
   }
 
+  public getExpansions(): string[] {
+    let expansions: any[] = this.lazyData.patches.map(p => {
```
Supamiu:
```suggestion
    const expansions: any[] = this.lazyData.patches.map(p => {
```

Never reassigned, prefer `const` over `let` (the linter won't accept it in CI).

apps/client/src/app/pages/inventory-optimizer/optimizations/consolidate-stacks.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  itemInExpansion(itemId: number, expansion: number): boolean {
+    // If the specified item does not have a patch associated with it, we do not filter it.
+    let itemPatch = this.lazyData.data.itemPatch[itemId];
+    if (itemPatch === null) {
+      return true;
+    }
+
+    // If no patch is defined for this patch that the item is assigned to, do not filter it.
+    let patch = this.getPatch(itemPatch);
```
Supamiu:
```suggestion
    const patch = this.getPatch(itemPatch);
```

apps/client/src/app/pages/inventory-optimizer/optimizations/consolidate-stacks.ts
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  @Memoized()
+  getPatch(id: number): any {
+    return this.lazyData.patches.find(p => p.ID === id);
+  }
+
+  itemInExpansion(itemId: number, expansion: number): boolean {
+    // If the specified item does not have a patch associated with it, we do not filter it.
+    let itemPatch = this.lazyData.data.itemPatch[itemId];
```
Supamiu:
```suggestion
    const itemPatch = this.lazyData.data.itemPatch[itemId];
```

apps/client/src/app/pages/inventory-optimizer/optimizations/inventory-optimizer.ts
```diff
@@ -38,6 +38,12 @@ export abstract class InventoryOptimizer {
     return this._getOptimization(item, inventory, data);
   }
 
+  /// Compare two InventoryItem objects to determine whether they consume the same slot in the same
+  /// container.
+  protected static inSameSlot(source: InventoryItem, target: InventoryItem): boolean {
```
Supamiu: Static methods are really not recommended in a Typescript context, especially for frontend with dependency injection. It makes it pretty much impossible to mock properly in a testing context, as typescript creates separate objects for static and dynamic contexts.

apps/client/src/app/pages/inventory-optimizer/inventory-optimizer/inventory-optimizer.component.ts
rcdailey: What's the benefit? I'm not familiar with it.

apps/client/src/app/pages/inventory-optimizer/inventory-optimizer/inventory-optimizer.component.ts
rcdailey: Will be in my next push.

apps/client/src/app/pages/inventory-optimizer/optimizations/inventory-optimizer.ts
rcdailey: It's a function that doesn't require object state, nor is it something that should be polymorphic. In C++ something like this would be a global or namespace-level free function. What's the philosophy and/or idiom in Typescript? For something like this I don't see how dependency injection is relevant.

Supamiu: The issue with static methods is that they can't be mocked at all in a TS context, that's why people tend to avoid static methods in TS. But this one can be fine as it's purely a helper.

apps/client/src/app/pages/inventory-optimizer/inventory-optimizer/inventory-optimizer.component.ts
Supamiu: Lodash provides small optimizations on various methods and that's why when they have an utility method that we can use, it's better to just use it instead of implementing our own solution.

apps/client/src/app/pages/inventory-optimizer/optimizations/inventory-optimizer.ts
Supamiu: However, if we keep this as static, it has to be the first field of the class, there's a lint rule for that.

apps/client/src/app/pages/inventory-optimizer/inventory-optimizer/inventory-optimizer.component.ts
```diff
@@ -120,14 +120,41 @@ export class InventoryOptimizerComponent {
...
     return item.retainerName || this.inventoryFacade.getContainerName(item.containerId);
   }
 
+  public getExpansions(): string[] {
+    const expansions: any[] = this.lazyData.patches.map(p => {
+      return { name: p.ExName, version: p.ExVersion }
+    });
+
+    return expansions.filter((value, index, array) => {
+      return index === array.findIndex(e => e.name == value.name);
```
Supamiu:
```suggestion
      return index === array.findIndex(e => e.name === value.name);
```

apps/client/src/app/pages/inventory-optimizer/optimizations/inventory-optimizer.ts
rcdailey: I'm not exactly fixated on it being static. That's a carry-over from C++ I learned related to OOP. If a method doesn't need state in a class, make it static. It's a way to promote separation of concerns / encapsulation. If that doesn't apply in Typescript I'm open to an alternative, just make a suggestion. If you have nothing I'll go ahead and move it to the top and we'll move on. Thanks!

apps/client/src/app/pages/inventory-optimizer/inventory-optimizer/inventory-optimizer.component.ts
rcdailey: Makes sense. The change will be in my next push.

apps/client/src/app/pages/inventory-optimizer/inventory-optimizer/inventory-optimizer.component.ts
rcdailey: I replaced this with `uniqBy` so the code mentioned here no longer exists.

PR #1390, opened 2020-02-25
Add fail cause localisation for users below min stats

latenightwreck:


latenightwreck:
I uhhh. probably shouldve done a rebase instead of a merge to catch up on commits >.>

Supamiu:
I'll merge but remember that you can just add the EN label, all the other ones will be handled by crowdin :)

PR #1413, opened 2020-02-28
fix: remove GST from universalis responses because it screwed with pr…

karashiiro:
…ofit calculations


karashiiro:
oh, that's out of date, big time

PR #1421, opened 2020-02-29
Fix consolidate stacks

rcdailey:
It was incorrectly identifying the NQ stack as the HQ stack. The
condition is reversed so that the HQ item is properly suggested.

Also reverted changes to files under `.vscode` to an earlier revision because recent changes were unintentional.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,27 +1,14 @@
 {
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
   "version": "0.2.0",
-  "configurations": [
-
-    {
-      "type": "electron",
-      "request": "launch",
-      "name": "Launch example",
-      "cwd": "${workspaceFolder}",
-      "appDir": "C:/Development/VSCodeProjects/CollabProjects/ffxiv-teamcraft/node_modules/electron-find/example2/main.js",
-      "sourceMaps": true,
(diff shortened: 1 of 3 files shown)


.vscode/launch.json
```diff
@@ -1,27 +1,14 @@
 {
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
```
Supamiu: You and @Nowbob will have to agree on this, because I see this change in your PRs everytime, youre VSCode adds these comments, Nowbob's removes it ^^ it's no big deal so it won't spot me from merging but if we could have a proper handling on this, it'd be better 😄

rcdailey: As I mentioned in my PR description, I reverted the state of the files under `.vscode` to the first commit I made that introduced them. After speaking with Bob about the issue, he acknowledged that his modifications to those files were unintentional while he was working on other issues. So I rolled the files back because as they were, they were broken.

VS Code is not mysteriously editing these files. It was all done by hand. I just want to reassure you that these changes are deterministic and intentional. I might clean up and remove the comment later. VS Code did generate that comment when the file was first created, but that's it. If anything I didn't remove it initially when I probably should have.

Nowbob:
I can confirm what rcdailey commented. My first PR (before I learned what I was doing) grabbed the broken file and I didn't catch it. I'll be making sure I pull the proper file and make sure it stays that way.

PR #1422, opened 2020-02-29
fix: Disable HQ setting works more thoroughly

rcdailey:
There are some edge cases in which this setting would not apply:

1. Creating a list from search results
2. Cloning another list

Now the setting for disabling HQ suggestions in lists works more
thoroughly.

Note that as part of this change, the "everyone" list property is also
more consistently handled.


rcdailey:
Please note I was not able to test this change locally on my machine. The UI requires the user be logged in to see the disable HQ checkboxes on a list. @Supamiu if you have advice on how to trick the localhost version of teamcraft to be logged in, I will test for you. Otherwise you can test it for me if you have the time.

PR #1423, opened 2020-02-29
fix(desktop): fixed retainer tracking KR/CN

Ra-Workspace:
Ok, It's really fixed.

I also learned how to build the app. It very works. 😄

### DIFF
--- a/apps/client/src/app/core/electron/machina.service.ts
+++ b/apps/client/src/app/core/electron/machina.service.ts
@@ -51,14 +51,11 @@ export class MachinaService {
       const borkedData = splitForCheck.findIndex((char) => {
         return char < ' ';
       });
-      if (borkedData > -1) {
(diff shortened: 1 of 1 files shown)


apps/client/src/app/core/electron/machina.service.ts
```diff
@@ -46,20 +46,10 @@ export class MachinaService {
...
-        name = name.substring(borkedData);
-      }
-      if (region === Region.Korea && environment.koreanGameVersion < 5.2) {
-        name = name.substring(4);
-      }
-      if (region === Region.China && environment.chineseGameVersion < 5.2) {
-        name = name.substring(4);
-      }
+      if ((region === Region.Korea && environment.koreanGameVersion < 5.2) || (region === Region.China && environment.chineseGameVersion < 5.2)) {
+        let uint8array: Uint8Array = new TextEncoder().encode(name);
```
Supamiu:
```suggestion
        const uint8array: Uint8Array = new TextEncoder().encode(name);
```

PR #1446, opened 2020-03-07
Share consumables with crafting rotation URLs

rcdailey:
feat: Share consumables with rotations

Now when including stats with crafting rotation links, the following
consumables are included:

* Selected Food
* Selected Medicine
* Up to two selected free company actions

This allows others that open your rotations via the link to see your
selected consumables. Previously the consumable stats were part of the
stats parameters which allowed others to add another food on top of it,
for example, effectively invalidating the simulation.


apps/client/src/app/pages/simulator/simulator.module.ts
```diff
@@ -1,4 +1,4 @@
-import { NgModule } from '@angular/core';
+import { NgModule, CUSTOM_ELEMENTS_SCHEMA } from '@angular/core';
```
Supamiu:
```suggestion
import { NgModule } from '@angular/core';
```

PR #1516, opened 2020-04-19
feat(inventory-optimizer): show materials not used in lists

corpozzy:
I'd rather not recalculate the list of materials every time a list is changed but couldn't find a better way to do it.

### DIFF
--- a/apps/client/src/app/pages/inventory-optimizer/optimizations/not-used-in-lists.ts
+++ b/apps/client/src/app/pages/inventory-optimizer/optimizations/not-used-in-lists.ts
@@ -1,0 +1,42 @@
+import { InventoryOptimizer } from './inventory-optimizer';
+import { InventoryItem } from '../../../model/user/inventory/inventory-item';
+import { UserInventory } from '../../../model/user/inventory/user-inventory';
+import { ListRow } from '../../../modules/list/model/list-row';
+import { LazyDataService } from '../../../core/data/lazy-data.service';
+import { ListsFacade } from '../../../modules/list/+state/lists.facade';
+import { Memoized } from '../../../core/decorators/memoized';
+import { OnDestroy } from '@angular/core';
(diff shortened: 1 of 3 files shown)


Supamiu:
Hello and thank you for your contribution 🙂 

While I understand the need for your feature, I don't think it belongs in the inventory optimizer, instead, it should be inside the list details page (which would solve the "update whenever a list is updated" problem too, and the memory leak).

The issue here is that having materials inside your inventories (inventory optimizer considers your inventory as well as your retainers' inventories).

I can review your code if you want to have feedback on it and how to make it more optimized, but I can't merge this in the inventory optimizer. It could, however, be adapted to fit inside the "inventory popup" inside the list details page (https://github.com/ffxiv-teamcraft/ffxiv-teamcraft/tree/staging/apps/client/src/app/pages/list-details/inventory-view)

apps/client/src/app/pages/inventory-optimizer/optimizations/not-used-in-lists.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export class NotUsedInLists extends InventoryOptimizer implements OnDestroy {
+  allMaterialsOnListsSubscription: Subscription;
+  allMaterialsOnLists: Set<number> = new Set();
+
+  constructor(private lazyData: LazyDataService, private listsFacade: ListsFacade) {
+    super();
+
+    this.allMaterialsOnListsSubscription = listsFacade.allListDetails$.subscribe(lists => {
+      this.allMaterialsOnLists = new Set(_.flatMap(lists, list => list.items.map(item => item.id)));
```
Supamiu, line 20: Instead of `flatMap`, I'd use `[].concat.apply([], lists.map(list => list.items.map(item => item.id)));`

Supamiu:
Closing this as we discusssed it and decided it was too niche for now, and it can live in a fork with no issues at all.

Issue #1517, opened 2020-04-28
bug: can't select the materia at the bottom in gear-set


Kaiay:
**Describe the bug**
gear-set can't select the materia at the bottom
**To Reproduce**
1. Go to 'gear-set'
2. Click on a list->a item
3. the 4ed and the last can't see(select) the materia at the bottom
**Expected behavior**
can see(select) the materia at the bottom
**Screenshots**
![image](https://user-images.githubusercontent.com/43386053/80462746-c05b4600-8969-11ea-8c32-7fadaafe20cc.png)

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: desktop app
 - Version v7.1.8

joseapintor:
Hey @Supamiu . Could I try taking a look at this? I worked on this by locally implementing a scrollbar.

PR #1518, opened 2020-04-29
Implement scrollbar to prevent drop-down items being inaccessible

joseapintor:
Fix for #1517 

Implemented a scrollbar to prevent drop-down items from being soft-locked by window/screen ratio.

### DIFF
--- a/apps/client/src/app/pages/gearset/materias-popup/materias-popup.component.html
+++ b/apps/client/src/app/pages/gearset/materias-popup/materias-popup.component.html
@@ -50,7 +50,7 @@
         <ul nz-menu>
           <li nz-menu-item (click)="resetMaterias(i)">{{'GEARSETS.No_materia' | translate}}</li>
           <li nz-submenu *ngFor="let row of materiaMenu" [nzTitle]="row.baseParamId | baseParamName | i18n">
-            <ul>
+            <ul style="width: 300px; height: 200px; overflow: auto">
               <li nz-menu-item
                   (click)="setMateria(i, materia)"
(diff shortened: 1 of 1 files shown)


Supamiu:
Hey, thank you for your PR, did you try to reproduce this using latest staging before implementing a fix? I tried to reproduce and couldn't do so, using web or desktop.

Your fix looks good-ish because the main issue is the fact that in-html style is really a bad practice in general, however sometimes it's required, when ::ng-deep didn't work properly, which happens on some specific cases.

Aside from that, 300px feels a bit arbitrary as it'll fix some cases but not others, I believe the main issue in this case is the cdk anchor, and it appears to be fixed with the recent angular 9 migration I did on staging (hence my question about reproductibility for you).

joseapintor:
Hey! Thanks for the quick reply. 

I just noticed my mistake, as I was trying to reproduce the issue on the latest release available (because the report said that it happened on the desktop 7.1.8 ver).

After trying again on the latest staging build (both web and app) it did indeed work perfectly on lots of different resolutions I tried yeah.

And I see the issue with the 300px, thanks for the explanation! I guess using something like `em` would have been better.

This is my first time having a go at OSS, had a lot of fun figuring out how to PR correctly, building everything myself, figuring out the code... so thanks for your supportive reply! I'd really to continue contributing, so I'll keep an eye out for any other difficulty: easy issues.

Supamiu:
I'm happy to see that you enjoyed that 🙂 I'll close this PR but be sure that your contributions, like every contributions, will be welcome.

PR #1548, opened 2020-07-06
fix(currency-spending): filter out shop listings with no npc

apmorton:
Garland contains old shop listings which no longer exist in game.
Many items have simply been moved under a different shop listing,
resulting in the currency spending page showing duplicate items.

You can see the incorrect behavior by looking at `Yellow Crafters' Scrip` currency, as shown below:
image

### DIFF
--- a/apps/client/src/app/pages/currency-spending/currency-spending/currency-spending.component.ts
+++ b/apps/client/src/app/pages/currency-spending/currency-spending/currency-spending.component.ts
@@ -86,7 +86,7 @@ export class CurrencySpendingComponent extends TeamcraftComponent {
         this.loading = true;
         return this.dataService.getItem(currency).pipe(
           map((item: ItemData) => {
-            return [].concat.apply([], item.item.tradeCurrency.map(entry => {
+            return [].concat.apply([], item.item.tradeCurrency.filter(entry => entry.npcs.length > 0).map(entry => {
               return entry.listings.map(listing => {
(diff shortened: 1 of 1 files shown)


apps/client/src/app/pages/currency-spending/currency-spending/currency-spending.component.ts
```diff
@@ -86,7 +86,7 @@ export class CurrencySpendingComponent extends TeamcraftComponent {
         this.loading = true;
         return this.dataService.getItem(currency).pipe(
           map((item: ItemData) => {
-            return [].concat.apply([], item.item.tradeCurrency.map(entry => {
+            return [].concat.apply([], item.item.tradeCurrency.filter(entry => entry.npcs.length).map(entry => {
```
Supamiu, line 89:
```suggestion
            return [].concat.apply([], item.item.tradeCurrency.filter(entry => entry.npcs.length > 0).map(entry => {
```

It's really nitpicking, but makes maintenance easier when i'll read this in six months and will take 5 more minutes to realize it's filtering on length > 0 because I won't read this properly for the first 2 times I'll read it 😄

Supamiu:
Hey, thank you for your PR 🙂 I'm glad you could find how to fix your issue without asking anyone ^^ (not that questions are bothering me, at all, but it shows me that the code isn't as bad as I could think it is).

Supamiu:
This will be included in the next update 👍 thanks again for the contribution 🎉

Issue #1458, opened 2020-03-12
ux: allow for mouse wheel up/down to edit amount of number input fields


Supamiu:
**Is your feature request related to a problem? Please describe.**
Not related to a problem.

**Describe the solution you'd like**
Mousewheel events on inputs increasing/decreasing amount like ingame.

**Describe alternatives you've considered**
focus => use arrow keys or click on arrow buttons inside the input.

**Additional context**
It could be a nice first directive to create, see http://blog.sodhanalibrary.com/2016/10/angular-2-directive-for-mouse-wheel.html for a pretty good example.

Once this directive is created, add it to all the `nz-input-number` elements in the app, and bind the wheel event emitters to the corresponding amount edition methods, based on the page it is in.

**Source**
https://discordapp.com/channels/355013337748209665/355110181148819467/687433079350100024

Nowbob:
Ooh this looks fun to try, but I can guarantee I will be slow on this one, if someone else *really* wants to take it over and finish it fast, please do.

ayyaruq:
This was completed with the release of 7.3.0, if you find latest Teamcraft doesn't do what you're expecting in regards to this feature, please leave a comment and we'll reopen this issue.

PR #1549, opened 2020-07-11
feat(allow-for-mouse-wheel): adding support for up and down

Jengines:
Hello,

First of all, this is my first contribution to this project, and the first time i tried to crontibuting to an Angular project. So, if i do some mistake do not hesitate to tell me.

And, you have probably already seen, i'm French, and my English is a little medium, I hope I can be understable.

**According to this issue #1458**

Mousewheel events on inputs increasing/decreasing amount like ingame.

What I have done : 
 - Adding a new directive to trigger an event for mouse events up and down (mouse-wheel.directive.ts)
 - Creating a new module (mouse-wheel.module.ts) to import him on every component we need to use mouse-wheel events
 - Binding the wheel events for following app pages : 
     - General -> List
     - Gathering -> Alarms -> Add custom alarm

What I have tried : 
 - Binding this directive to item-picker input field, but this field isn't a `<nz-input-number>` and when I have applied this directive to the input component the behaviour wasn't the same 

What I ask to you : 
 - Can you review my code and tell me if i have done something bad ? Every minor remark is usefull to me
 - If we want to continue this feature, all `<nz-input-number>` should have this directive ? If yes, my approach looks good ?

When coding this feature I encoutered some suspect behaviours, not due to my added code, but already exist in TC 7.2.8, following : 
 - When updating the number of item on General -> List, before, we can specify a negative number, i have removed this possibility and added a minimal limit to 0, but, when we set 0 to this field, the item disappear from the list, is that behavior good ? It's look weird to me.
 - Their is no maximum limit for item on General -> List, so we can go to an Infinity number of items an the app will crash, should we apply a maximum number of craftable items ?

Thank's you all


apps/client/src/app/core/event/mouse-wheel/mouse-wheel.directive.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  @HostListener('DOMMouseScroll', ['$event']) onMouseWheelFirefox(event: any) {
+    this.mouseWheelFunc(event);
+  }
+
+  @HostListener('onmousewheel', ['$event']) onMouseWheelIE(event: any) {
+    this.mouseWheelFunc(event);
+  }
+
+  mouseWheelFunc(eventUi: any) {
```
Supamiu, line 20:
```suggestion
  mouseWheelFunc(eventUi: any): void {
```

Even if we don't know the type of the event, we can type the return type as void, making it easier to read as we do'nt have to look for a return 🙂

Supamiu:
Hey, thank you for your nice first PR, I'm happy to say it looks really good for a first contribution 👍 .

I'm going to reply your points one by one, but keep in mind setting up a vocal discussion on discord is totally possible for me, and would probably make things easier with screen share etc. Would make things even easier considering we both have the same native language 😉 .

> If we want to continue this feature, all <nz-input-number> should have this directive ? If yes, my approach looks good ?
Yes I think nz-input-number should contain this directive, but for that there's two solutions:

 1. Paste the directive everywhere we're using nz-input-number, to add the behavior.
 2. Create our own number input component, using nz-input-number and adding common logic like the wheel behavior.

> When updating the number of item on General -> List, before, we can specify a negative number, i have removed this possibility and added a minimal limit to 0, but, when we set 0 to this field, the item disappear from the list, is that behavior good ? It's look weird to me.

This was intended because when you set the amount to 0, it sets all the other amounts to 0, and after each addition operation on a list, it's cleaned, meaning that every item at amount: 0 is removed. Maybe we want to avoid this when it comes from mouse wheel event? I don't know.

> Their is no maximum limit for item on General -> List, so we can go to an Infinity number of items an the app will crash, should we apply a maximum number of craftable items ?

I'm surprised it crashes at all, but I think it's only related to the nature of Infinity. In beta env, we created lists with insane amounts like 10M+ of an item, and it worked like a charm. So unless the user can input Infinity, I think we're safe here 👍 .

Jengines:
Thank you for your answers, I have commited your suggestion.

We can discuss on Discord for the further of this request, and see what I can do !

Jengines:
According to @Nowbob codebase (see https://github.com/Nowbob/ffxiv-teamcraft/compare/staging...Nowbob:feat/mouse-wheel-number-inputs)

I have review my PR with the following changes :
 - The directive mouseWheel is now injected on the core module (so we don't need to add her on all modules we want)
 - Each function applied to mouse wheel up and down are now more reusable (we pass the current formControlName and the other parameters we need)
 - Some boundaries have been added to prevent scroll up onto negative numbers, and scroll up onto infinity number where it doesn't have sense

The pages of the application where this feature is applied are :
 - Gathering -> Alarms -> Add custom alarm (Spawn, Duration, Slot, X and Y fields)
 - General -> Lists -> (Click on the list to expand) Increment/Decrement number of each item
 - General -> Search -> Increment/Decrement number of each item on the result page
 - General -> Search -> Filters 
							-> Item (Item level / Required level / Recipe level)
							-> Recipe (Item level / Required level / Recipe level)
							-> Instance (Level)
						    -> Levequest (Required level)
							-> Action (Level)
							-> Trait (Level)
 - General -> Crafting rotations -> "Any rotation" -> Recipe configuration panel (rlvl / lvl / progress / quality / durability / Suggested Craftmanship / Suggested Control)
 - General -> Crafting rotations -> "Any rotation" -> Share with the community (Min. Craftmanship, Min.Control, Min. CP)
 
 
List of all boundaries added with this feature
 - Item added onto a list cannot be minus than 1 and more than 999
 - Coordinates X and Y on custom alarm cannot exceed 99
 - maxiLvl on search filter cannot exceed 999
 - maxStats on search filter cannot exceed 99999
 - All fields on panel "Recipe configuration" for rotation cannot be minus than 1
 - Field lvl on panel "Recipe configuration" for rotation cannot exceed 80 (actual max lvl for Shadowbringers)
 
Hope this first changes can make the application a little more user friendly, and we can loke after to added this on other fields.

Thank's for the review !

apps/client/src/app/modules/list/list-panel/list-panel.component.ts
```diff
@@ -38,6 +38,10 @@ import { TeamcraftComponent } from '../../../core/component/teamcraft-component'
 })
 export class ListPanelComponent extends TeamcraftComponent {
 
+  //Bounds for number of a single item in a list
+  maxAmount = 999;
```
Supamiu, line 42: 999 is way too low, In my opinion, if you want a maximum, you have to make it at least 10^9 because technically, you can even list gils in lists.

Even without considering gils, a lot of lists actually have 1000+ of a single item listed, for stuff like pre-ultimate consumables prep.

apps/client/src/app/pages/search/search-result/search-result.component.ts
```diff
@@ -11,6 +11,10 @@ import { HtmlToolsService } from '../../../core/tools/html-tools.service';
 })
 export class SearchResultComponent {
 
+  //Bounds for number of a single item to be added to list
+  maxAmount = 999;
```
Supamiu, line 15: Same as for the list panel maxAmount property (might want to put these inside a global configuration file)

apps/client/src/app/pages/search/search/search.component.ts
```diff
@@ -755,4 +760,18 @@ export class SearchComponent implements OnInit {
   trackByItem(index: number, item: SearchResult): number {
     return +item.itemId;
   }
+
+  public adjust(form: string, prop: string, amount: number, min: number, max: number, arrayName?: string, arrayIndex?: number): void {
```
Supamiu, line 764:
```suggestion
  public adjust(form: keyof SearchComponent, prop: string, amount: number, min: number, max: number, arrayName?: string, arrayIndex?: number): void {
```

This small change makes sure you're giving a form name that exists.

Jengines: You mean prop: keyof SerchComponent no ?

apps/client/src/app/modules/list/list-panel/list-panel.component.ts
Jengines: Should I apply a limit to 10^9, or find a way to not apply any max limit for this field ?

apps/client/src/app/pages/search/search/search.component.ts
Supamiu: no, form, because form is a member of SearchComponent

apps/client/src/app/modules/list/list-panel/list-panel.component.ts
Supamiu: How did you manage to reach infinity using scroll wheel? Because I am really wondering how it is even possible, considering chrome handles numbers > 10^32

apps/client/src/app/pages/search/search/search.component.ts
Jengines: Or not, but SearchComponent is not the good type, with FormGroup it work's, don't know if it's what you think about

apps/client/src/app/modules/list/list-panel/list-panel.component.ts
Jengines: I have never reached Inifinity using scroll wheel (yes, it's infinity so .. !), but i put this max bounds to be coherent with the other bounds on search-result.component.ts. I approve to put a limit to prevent user creating list with foolish amounts of items on it. You told me that test on Beta serv with millions of items doesn't create any crash of the application, so maybe we can just remove the max limit !

apps/client/src/app/pages/search/search/search.component.ts
Supamiu: the form you're passing is a name, the name of the form inside of the SearchComponent component, which is why it's this type.

For instance, one of the values used is `instanceFiltersForm`, which is a property of SearchComponent.

PR #1568, opened 2020-08-04
fix(gathering): fix redundant weather content if weather is not present

Airex:


Supamiu:
Thank you for this nice first PR 😃 , it'll be included inside the 7.3 update, hopefully released this tuesday or wednesday.

PR #1593, opened 2020-08-19
feat(pricing): better world name displaying for cn servers

zhyupe:
Add l12n to CN server names in marketboard and pricing mode (display Chinese words like '沃仙曦染' instead of pinyin 'WoXianXiRan'):

image
image


apps/client/src/app/pipes/pipes/world-name.pipe.ts
```diff
@@ -0,0 +1,18 @@
...
+@Pipe({
+  name: 'worldName'
+})
+export class WorldNamePipe implements PipeTransform {
+
+  constructor(private data: LocalizedDataService) {
+  }
+
+  transform(world: string): I18nName {
+    console.log('world', world)
```
Supamiu, line 14: Oops 😉

apps/client/src/app/modules/pricing/pricing/pricing.component.ts
```diff
@@ -226,12 +226,12 @@ export class PricingComponent implements AfterViewInit {
         const amount = this.pricingService.getAmount(list.$key, row, false);
         let priceString: string;
         if (price.hq > 0 && amount.hq > 0) {
-          priceString = `${price.hq.toLocaleString()}gil x${amount.hq}(HQ) (${price.hqServer})`;
+          priceString = `${price.hq.toLocaleString()}gil x${amount.hq}(HQ) (${this.getWorldName(price.hqServer)})`;
```
Supamiu, line 229: Why not use the pipe instead?

Supamiu (changes requested):

apps/client/src/app/pipes/pipes/world-name.pipe.ts
zhyupe: Oops, fixed

apps/client/src/app/modules/pricing/pricing/pricing.component.ts
zhyupe: Referencing to codes like `this.i18n.getName(this.l12n.getItem(row.id))`, I think it's reasonable (not perfect, sure). Also calling pipes will result in 2 more dependency injections.

Supamiu (approved):

PR #1599, opened 2020-08-24
Feat/add exclusion filter

### LINKED ISSUE #1460: feat(community-lists): add tags exclusion filter
**Is your feature request related to a problem? Please describe.**
When browsing community lists to, for instance, look for gear crafted by LTW, you'll enter LTW and Gear tags, but then you're flooded with bazillions of lists because a lot of lists match these two tags.

**Describe the solution you'd like**
A way to input tags that would be excluded from the search results.

**Describe alternatives you've considered**
*N/A*

**Additional context**
The filter can be a simple clone of the current tags input, placed below it.

TS-wise, it has to be a `map` operator added to the results observable, as first parameter of this pipe call : 
https://github.com/ffxiv-teamcraft/ffxiv-teamcraft/blob/8fa0fb5642d5762095eef36130c384351a8bb3bf/apps/client/src/app/pages/community-lists/community-lists/community-lists.component.ts#L73

As the exclusion cannot be done inside firestore (because we already use array-contains matcher), the filtering has to be done client-side, inside this new `map` operator.

**Source**
https://discordapp.com/channels/355013337748209665/355110181148819467/682299847319093353

bryce0110:
Resolves #1460 

Simple UI, just a copy of the tags filter placed below it. The exclusion filter is included as a query string, similar to the tags filter.

It checks if at least one of the lists' tags is also in the exclude filter, and if it is then it will be filtered out of the displayed lists.


bryce0110:
Not sure why it included so many commits there. I might've done some weird git stuff on my end lol.

PR #1604, opened 2020-08-26
Perf: Fish data loading improvements + lazy data deferral

emily-curry:
## Motivation

As someone that uses the fish pages of the db often, it's occasionally not very responsive. The single query to load all the fish data times out, which causes it to be retried (which may result in further timeouts), or otherwise cause the page to be unresponsive. My goal is to improve the time-to-interaction for each fish db page load (and eventually do the same for fishing spots).

## Implementation Approach

I'm opening this PR as a draft to collect feedback on the following approaches, as it's getting kind of large and I want to be sure it's in-line with the direction of the project.

### Query Splitting

Splitting the single gql query into multiple parallel queries resolved the timeout issues I had been seeing. This also allows us to break the queries out into more manageable chunks, which can then be typed and used for observable flows a little more ergonomically. [`fish-data.gql.ts`](apps/client/src/app/pages/db/service/fish-data.gql.ts) contains the typed queries, `FishDataService` contains methods for creating observables from those queries, and `FishDataContextService` has a concept of a "current fish" and a "current spot" which it uses to define observables for loading information about the current fish/spot. This service is what's eventually used by the component, keeping the data loading concerns isolated from the view layer. Also, it's all type-safe now!

I'm not super familiar with firebase, but if functions are billed by cpu-minutes then I'm hopeful that it's cheaper to execute 6 concurrent queries that all succeed than one that times out and is retried (if not, this can be tuned to make less requests).

### Apollo Caching

The existing functionality had the gql query being made explicitly not use caching, so that makes me a little worried about this change, but I've removed that so that Apollo always checks the cache first. It'd be possible to manually create a data store with a bunch of mergeScans, but I figured Apollo was the best tool for the job of managing the caching of this data.

This change keeps all loaded fish in-memory, which means that when navigating between db pages, fish that you've already viewed in the current session will reload near-immediately (this is something I do very often, so this is a huge usability boost).

### Lazy Data Deferral

This is probably the biggest change, as it's introducing some architectural modifications (or at least a path to them). Basically, while refactoring the fish data to be more observable-driven, I realized some of the view data comes from the `LazyDataService`, which loads a lot of data on app init, which the fish component then accesses either directly or though `LocalizedDataService`. This seemed at odds with the observable data flows, and also required the user to load a _lot_ of data up front. To address this, I've created a `LocalizedLazyDataService`, which serves the same function, but returns observables that will load the data instead. This means that we can define _how_ data is loaded all across the app without worrying about weird timing issues (which, my guess would be is what loading the bulk of it up front is addressing), but _when_ the required data is loaded can be deferred until something actually subscribes. Of course, if there are "hot paths" that we should pre-load, that should be relatively trivial to implement as well.

## Functional Differences

I tried to keep the functionality of the page the exact same. The following behaviors are different as a result of this PR, however:

- Removed "refresh" button on fish page, as it would just hit the cache. Will look into apollo cache invalidation, but I figured this data changed infrequently enough for it to not be super important.

## TODO

Things that should be completed before this is ready:
- [x] Inline docs
- [x] Update `I18nPipe` to also accept the new `I18nNameLazy` values
- [x] Finish implementation of LocalizedLazyDataService
- [x] Add some kind of loading UI to cards on fish pages while data is loading.


apps/client/src/app/core/data/lazy-data.service.ts
```diff
@@ -62,21 +61,30 @@ export class LazyDataService {
   public data: LazyData;
   public data$: ReplaySubject<LazyData> = new ReplaySubject<LazyData>();
 
+  public readonly fishingSpots$ = this.lazyDataProvider.getLazyData('fishingSpots');
```
emily-curry, line 64: Besides the addition of this line, changes at L250 to accommodate the lazy file list structure change, this file is all changes made by prettier.

apps/client/src/app/core/data/localized-data.service.ts
```diff
@@ -1,26 +1,25 @@
...
+import { mapIds } from './sources/map-ids';
 import { tripleTriadRules } from './sources/triple-triad-rules';
 import { zhActions } from './sources/zh-actions';
 import { zhWorlds } from './sources/zh-worlds';
-import { ExtendedLanguageKeys } from './extended-language-keys';
-import { LazyData } from './lazy-data';
-import { Trait } from '../../pages/db/model/trait/trait';
 
-@Injectable()
+@Injectable({ providedIn: 'root' })
```
emily-curry, line 18: Besides this line, this file is also basically all prettier changes.

apps/client/src/app/pages/db/db.module.ts
```diff
@@ -267,8 +265,8 @@ const routes: Routes = [
 
     NgZorroAntdModule,
     UserAvatarModule,
-    QuickSearchModule
-  ]
+    QuickSearchModule,
+  ],
+  providers: [FishDataService, FishContextService, ...Object.values(FishGQLProviders)],
```
emily-curry, line 270: Besides this line, also all prettier.

apps/client/src/app/core/data/localized-lazy-data.service.ts
```diff
@@ -0,0 +1,463 @@
...
+
+  public getWorldName(world: string): I18nNameLazy {
+    const i18nName: I18nNameLazy = {
+      fr: of(world),
+      en: of(world),
+      de: of(world),
+      ja: of(world),
+      zh: of(zhWorlds[world]),
+      ko: of(undefined),
+      ru: of(undefined),
```
Supamiu, line 84: Because this is about world name label, I'd rather not show undefined, but get the base value instead.

```suggestion
      ko: of(world),
      ru: of(world),
```

apps/client/src/app/core/data/localized-lazy-data.service.ts
```diff
@@ -0,0 +1,463 @@
...
+import { ExtendedLanguageKeys } from './extended-language-keys';
+import { Language } from './language';
+import { LazyDataKey } from './lazy-data';
+import { LazyDataProviderService } from './lazy-data-provider.service';
+import { lazyFilesList } from './lazy-files-list';
+import { mapIds } from './sources/map-ids';
+import { zhWorlds } from './sources/zh-worlds';
+
+@Injectable({ providedIn: 'root' })
+export class LocalizedLazyDataService {
```
Supamiu, line 16: Maybe this should implement LazyDataService, to make sure they expose the same methods over time.

apps/client/src/app/pages/db/fish/fish.component.html
```diff
@@ -1,269 +1,287 @@
...
             </div>
-            <div class="details-block no-icon">
-              <div class="detail-name">
-                {{'DB.FISH.Min_size' | translate}}
-              </div>
-              <div class="detail-value">
-                {{(data.minSize / 10) | number:'1.0-1'}} ilm
-              </div>
+            <div class="detail-value">
+              {{ (fishCtx.statisticsByFish$ | async)?.data?.stats.aggregate.min.gathering | number: '1.0-1' }}
```
Supamiu, line 11: I see this `statisticsByFish$` used a lot, it'd be better to have it aliased inside a wrapping `ng-container` imo, then wrapping each card with its own [nz-spin](https://ng.ant.design/components/spin/en)

Supamiu:
Hey, thank you for this really nice first contribution, having someone who wants to dive this deep inside the code without having to ask me anything about it is impressive and really enjoyable 👍 

Your refactoring to the fish data loading is really impressive and I like to see how it's working and how better it'll be for the end user experience, especially once each card is able to use [`nzLoading`](https://ng.ant.design/components/card/en#nz-card ) to show its loading state.

Regarding your refactoring, it's something I wanted to do for a long time, but there's some points you have to consider, that made me have a hard time figuring out a proper solution (but I think you'll figure it out):

 - Originally, LazyDataService is here to preload stuff so the pipes like `I18nName` can be pure, meaning that the change detector will be lazy and just call them once per component life. This is kinda of a hack becaus etechnically, since the data is loaded from async stuff, it's not pure, but since we don't show the app if the data isn't loaded, it's pure. Making these pipes impure would mean having a potential decrease in ap speed, that has to be monitored.

 - A lot of services are using `LazyDataService.data` in a sync way to make things easier and faster, I'm not sure you want to convert all these to Observables.

While I love the perspective of having everything lazy loaded _for real_, it's a really big thing and having fish data loaded in fragments is already a big improvement, don't feel forced to make everything perfect regarding lazy data.

But if you want to, I'll be happy to help you 👍

apps/client/src/app/core/data/lazy-data-provider.service.ts
```diff
@@ -0,0 +1,46 @@
...
+  /**
+   * Gets an observable that will resolve lazy data for the specified key.
+   * For each key, only a single `ReplaySubject`-like observable will ever be created.
+   * This ensures that each key will only ever be loaded once, but that loading will be
+   * defered until the observable is actually subscribed to.
+   *
+   * @param key The key of the lazy data to retrieve.
+   * @returns An observable that will emit the lazy data associated with the specified key, then completes.
+   */
+  public readonly getLazyData = <T extends LazyDataKey>(key: T): Observable<LazyData[T]> => {
```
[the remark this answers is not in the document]
emily-curry: It is just me being defensive with functions that might be used as arguments to other functions. Typescript doesn't implicitly defend against unbound `this` (unless the angular bundler is doing some magic on top of it that I'm unaware of).

apps/client/src/app/core/data/localized-lazy-data.service.ts
```diff
@@ -0,0 +1,463 @@
...
+      ru: this.getFallbackResolver(ruKey, key, accessor, 'ru'),
+      zh: this.getFallbackResolver(zhKey, key, accessor, 'zh'),
+    };
+  }
+
+  private guessExtendedLanguageKeys(key: LazyDataKey): ExtendedLanguageKeys {
+    return {
+      zhKey: this.guessExtendedLanguageKey('zh', key),
+      koKey: this.guessExtendedLanguageKey('ko', key),
+      ruKey: this.guessExtendedLanguageKey('ko', key),
```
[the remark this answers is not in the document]
emily-curry: Yep! Oops.

apps/client/src/app/core/data/localized-lazy-data.service.ts
emily-curry: Hm, I wonder if that might be a little awkward, since the methods have different signatures? Do you think they should all be suffixed with something, maybe (like `getItemLazy(id: number): Observable<I18nNameLazy>`)?

apps/client/src/app/pages/db/fish/fish.component.html
emily-curry: Definitely! I'll address that as part of the UI cleanup.

emily-curry:
Hey, I'm glad to hear it's a welcome change! It can sometimes go either way with these larger PRs 😅

I'll continue on with polishing this up, then. For this PR, I can finish displaying async state in the UI (thanks for linking the components, I'm totally unfamiliar with this component library), and finish all the TODOs in the LocalizedLazyDataService.

> A lot of services are using LazyDataService.data in a sync way to make things easier and faster, I'm not sure you want to convert all these to Observables.

I do agree that having everything really be lazy loaded is a pretty big task (and probably outside the scope of what I'm doing here). So more than anything, this could perhaps be considered a proof of concept for lazy loading that has an incremental adoption path? Hopefully. if there are two services with similar interfaces, it should be not too terrible to swap out the sync one for the async one. But unfortunately, we won't really see the benefits of it until we can stop loading data on app init.

So yeah, on one hand it duplicates functionality (which in itself might introduce new complexity), but on the other hand, I think it's probably a step in the right direction! I'll let y'all make the final call on that one though 🎉

> Originally, LazyDataService is here to preload stuff so the pipes like I18nName can be pure, meaning that the change detector will be lazy and just call them once per component life. This is kinda of a hack becaus etechnically, since the data is loaded from async stuff, it's not pure, but since we don't show the app if the data isn't loaded, it's pure. Making these pipes impure would mean having a potential decrease in ap speed, that has to be monitored.

This is definitely something to keep an eye on! I don't want to optimize too heavily in advance (since I don't anticipate it being too bad, the `async` pipe itself is impure, after all), but if it becomes noticeable anywhere, using `ChangeDetection.OnPush` in those update-heavy components is a pretty easy solution.

apps/client/src/app/core/data/localized-lazy-data.service.ts
Supamiu: Hm indeed, This means that either we create an interface with a generic type for returns but it'll be a pain, or we let them as they are right now.

Yeah the latter is the better I think here 🙂

apps/client/src/app/core/data/lazy-data-provider.service.ts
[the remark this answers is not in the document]
Supamiu: Ok I see, I think it's better to have them as functions right now, because there's pretty much nothing that takes a function as parameter in TC as I'm always going for rxjs over callbacks, and the only functions created on the fly are pure.

Supamiu:
> Hopefully. if there are two services with similar interfaces, it should be not too terrible to swap out the sync one for the async one. But unfortunately, we won't really see the benefits of it until we can stop loading data on app init.

Sadly it is harder than I'd like it to be, because in order to swap to full lazy, stuff like `this.lazyData.data.recipes` has to be changed to either `subscribe` to the recipes observable (which is something I don't like at all) or refactor the logic to provide the result inside an observable, meaning a lot of logic refactoring.

However, I'm not opposed to this refactoring, just that it's a super big chunk of work 😄 

Also, something to keep in mind is the main data file: [extracts.json](https://github.com/ffxiv-teamcraft/ffxiv-teamcraft/blob/staging/apps/client/src/assets/extracts.json). This file is loaded at boot time because it's required to load lists properly.

Before TC 6.0, lists had informations about every single item stored inside of them, this created issues with firestore as it's limited to 1MB documents and some items, such as crystals, have a lot of informations tied to them. I decided to refactor this and compile all the informations about every single item inside one file: extracts.json. This file is required everywhere and it used for lists, which are the base of Teamcraft, and loaded at boot time. This file is always 10MB+ but like every other data asset files, it's cached for a year (because it has a hash inside its name).

If you want to benchmark the solution, switch the itemName pipe for the new async version inside [app-item-row](https://github.com/ffxiv-teamcraft/ffxiv-teamcraft/blob/staging/apps/client/src/app/modules/list/item/item-row/item-row.component.html) and open a 100+ items list, it should be smooth after it loaded, else it means that the change detector is running too hard.

apps/client/src/app/pages/db/fish/fish-baits/fish-baits.component.ts
```diff
@@ -0,0 +1,40 @@
...
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class FishBaitsComponent {
+  public readonly spotIdFilter$ = this.fishCtx.spotId$.pipe(map((spotId) => spotId ?? -1));
+
+  public readonly loading$ = this.fishCtx.hoursByFish$.pipe(map((res) => res.loading));
+
+  public readonly baitsChartData$ = this.fishCtx.baitsByFish$.pipe(
+    switchMap((res) => {
+      if (!res.data) return of(undefined);
```
Supamiu, line 23: You can return an empty array here so you don't need to use `|| []` inside the template.
```suggestion
      if (!res.data) return of([]);
```

apps/client/src/app/pages/db/fish/fish-hooksets/fish-hooksets.component.ts
```diff
@@ -0,0 +1,43 @@
...
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class FishHooksetsComponent {
+  public readonly spotIdFilter$ = this.fishCtx.spotId$.pipe(map((spotId) => spotId ?? -1));
+
+  public readonly loading$ = combineLatest([this.fishCtx.hooksetsByFish$, this.fishCtx.tugsByFish$]).pipe(map(([hook, tug]) => hook.loading || tug.loading));
+
+  public readonly hooksets$ = this.fishCtx.hooksetsByFish$.pipe(
+    map((res) => {
+      if (!res.data) return undefined;
```
Supamiu, line 20: You can return an empty array here so you don't need to use `|| []` inside the template.
```suggestion
      if (!res.data) return [];
```

apps/client/src/app/pages/db/fish/fish-hooksets/fish-hooksets.component.ts
```diff
@@ -0,0 +1,43 @@
...
+        .map((entry) => ({
+          hookset: entry.id === 1 ? 4103 : 4179,
+          percent: (100 * entry.occurrences) / res.data.total,
+        }));
+    })
+  );
+
+  public readonly tugs$ = this.fishCtx.tugsByFish$.pipe(
+    map((res) => {
+      if (!res.data) return undefined;
```
Supamiu, line 32: You can return an empty array here so you don't need to use `|| []` inside the template.
```suggestion
      if (!res.data) return [];
```

Supamiu:
After looking at your changes again, I started digging for apollo cache handling, and I found [this article](https://medium.com/@martinseanhunt/how-to-invalidate-cached-data-in-apollo-and-handle-updating-paginated-queries-379e4b9e4698) about apollo cache invalidation on-demand, maybe it'll be helpful to you 👍

emily-curry:
> If you want to benchmark the solution, switch the itemName pipe for the new async version inside [app-item-row](https://github.com/ffxiv-teamcraft/ffxiv-teamcraft/blob/staging/apps/client/src/app/modules/list/item/item-row/item-row.component.html) and open a 100+ items list, it should be smooth after it loaded, else it means that the change detector is running too hard.

Thanks for this suggestion! I've been testing this by taking pipes like the `itemName` pipe that return `I18nName` values, and changing them to return `I18nNameLazy` values. The `i18n` pipe now handles both pretty seamlessly. I've still got some more testing to do (which I am doing on [this branch](https://github.com/emily-curry/ffxiv-teamcraft/compare/feat/fish-data-loading...emily-curry:feat/lazy-pipe-swap?expand=1) to keep those changes separate), but so far it's looking really good. Like, no noticeable impact at all with 126 items.

apps/client/src/app/core/data/lazy-data.ts
```diff
@@ -1,4 +1,6 @@
-export interface LazyData { 
+export type LazyDataKey = keyof LazyData;
```
Supamiu, line 1: I know it might seem overkill, but having a file for this type will be needed, because this file is generated on the fly by the lazy data updater script.

Supamiu (commented):
I see you also moved the company actions file, make sure you also change its usages inside the simulator component and the rotation service 👍

apps/client/src/app/core/data/lazy-data.ts
emily-curry: I actually included this line in the generator script itself.

Supamiu: Oh? Didn't see it when I checked it, my bad ^^

emily-curry:
> I see you also moved the company actions file, make sure you also change its usages inside the simulator component and the rotation service 👍
So, interestingly, that table contained both a ts and json file. The ts I left untouched, and that is what the simulator references. The json seemed unused beside the LocalizedDataService, so I thought it safe to move.

Also, if you're alright with this approach, I think I'll remove all the bits of `LocalizedLazyDataService` that are unimplemented, and include them in a different PR that's more focused on the lazy data, so that we can get the fish page changes in. I mostly just wanted to prove out the approach and make sure it was performant.

Supamiu:
> So, interestingly, that folder contained both a ts and json file. The ts I left untouched, and that is what the simulator references. The json seemed unused beside the LocalizedDataService, so I thought it safe to move.

Yeah I think some cleanup is still needed for some of the data files that should be lazy 😄 

> Also, if you're alright with this approach, I think I'll remove all the bits of LocalizedLazyDataService that are unimplemented, and include them in a different PR that's more focused on the lazy data, so that we can get the fish page changes in. I mostly just wanted to prove out the approach and make sure it was performant.

I am ok with that. 

Do you want to merge the PR as it is right now? I saw you removed the draft state and if we merge it soon enough, it can be tested properly since next big update is going to be in a week (5.31).

Supamiu (approved):

Supamiu:
I think the most complicated part will be to handle pages that are actually using lazy data, such as recipe finder, db pages (especially for display where it shows name in every language). I'd love to discuss it and find proper solutions tho.

I'm starting to think maybe we should create an issue or do you prefer talking about it inside the next PR?

emily-curry:
> Do you want to merge the PR as it is right now? I saw you removed the draft state and if we merge it soon enough, it can be tested properly since next big update is going to be in a week (5.31).


Yep! It's ready to go. We've mostly been talking about lazy data, but I am excited about how the fish page feels now. Caching made the biggest difference in terms of UX, but it's also very nice (in my opinion) to see the individual cards load as the data for them comes in.

>I think the most complicated part will be to handle pages that are actually using lazy data, such as recipe finder, db pages (especially for display where it shows name in every language). I'd love to discuss it and find proper solutions tho.

I'm starting to think maybe we should create an issue or do you prefer talking about it inside the next PR?


I think an issue for identifying everywhere that might be impacted and tracking progress would be a great idea, since this is likely to span multiple PRs.

PR #1623, opened 2020-09-10
Feat/json data export

iturtleman:
- Various minor changes that vastly improves the render performance of list-view's copy buttons
- Add ability to export  a list as JSON with some details (for use in other tooling)

Note: this is my first time making something in Angular. Please let me know if there's something I'm blatantly doing wrong!

Created a list from a random set of high level items from GarlandTools and the output looks like attached

[listCopy.txt](https://github.com/ffxiv-teamcraft/ffxiv-teamcraft/files/5201889/listCopy.txt)


apps/client/src/app/modules/list/list-details-panel/list-details-panel.component.ts
```diff
@@ -91,13 +99,24 @@ export class ListDetailsPanelComponent implements OnChanges, OnInit {
...
+    private eorzeaFacade: EorzeaFacade, private alarmsFacade: AlarmsFacade,
+    public settings: SettingsService, private lazyData: LazyDataService,
+    private authFacade: AuthFacade,
+    private gt: GarlandToolsService,
+    private _clipboardService: ClipboardService
+  ) {
+    //hold onto server name once resolved
+    this.authFacade.mainCharacter$.pipe(
+      map(char => char.Server)
+    ).subscribe(this.server$);
```
Supamiu, line 119: Instead of doing this, you can just have `server$` be an `Observable` and make it 

```ts
    this.authFacade.mainCharacter$.pipe(
      map(char => char.Server)
    )
```

The general rule is to avoid subscription, because unhandled subscriptions = memory leaks.

iturtleman:
I believe this may fail while in `Anonymous` mode. I will double check this / fix tomorrow.

apps/client/src/app/modules/list/list-details-panel/list-details-panel.component.ts
```diff
@@ -437,6 +465,11 @@ export class ListDetailsPanelComponent implements OnChanges, OnInit {
     });
   }
 
+  public copyTextExport() {
+    if (this._clipboardService.copyFromContent(this.getTextExport()))
+      this.textCopied();
+  }
```
Supamiu, line 471: Using the [cdk directive](https://material.angular.io/cdk/clipboard/overview) here would be better, as it's doing exactly this but it's maintained without us having to do anything.

apps/client/src/app/modules/list/list-details-panel/list-details-panel.component.ts
```diff
@@ -455,6 +488,30 @@ export class ListDetailsPanelComponent implements OnChanges, OnInit {
...
+      }, []);
+    } else {
+      rows = this.displayRow.rows;
+    };
+
+    if (this._clipboardService.copyFromContent(
+      JSON.stringify(this.getSerializedRowData(rows))//,null,2)
+    ))
+      this.jsonCopied();
+  }
```
Supamiu, line 505: Using the [cdk directive](https://material.angular.io/cdk/clipboard/overview) here would be better, as it's doing exactly this but it's maintained without us having to do anything.

apps/client/src/app/modules/list/list-details-panel/list-details-panel.component.ts
```diff
@@ -455,6 +488,30 @@ export class ListDetailsPanelComponent implements OnChanges, OnInit {
...
+
+  private getSerializedRowData(rows: ListRow[]): any {
+    return this.serializationHelper.getJsonExport(UniversalisService.GetDCFromServerName(this.lazyData.datacenters, this.server$.getValue()), this.server$.getValue(), rows);
```
Supamiu, line 508: You can't use a behaviorSubject just to have a sync value, if you want to be able to use the server as value, make it a function parameter and inject it from the template using the async pipe.

Supamiu (changes requested):

iturtleman:
can have some linting conflicts with https://github.com/ffxiv-teamcraft/ffxiv-teamcraft/pull/1624/files

Supamiu (approved):

iturtleman:
Tested with reading in c# (rename to .linq)
[GatheringHelper.linq.txt](https://github.com/ffxiv-teamcraft/ffxiv-teamcraft/files/5212878/GatheringHelper.linq.txt)

iturtleman:
A few missing things still 
- actual final list of items
- human readable instances info

apps/client/src/app/modules/list/data/ListRowSerializationHelper.ts
```diff
@@ -0,0 +1,268 @@
...
+    return getItemSource<T>(row, type, isObject);
+  }
+
+  constructor(
+    private i18nTools: I18nToolsService,
+    private l12n: LocalizedDataService,
+    private gt: GarlandToolsService,
+  ) {
+  }
+  public applyItemName(obj) {
```
Supamiu, line 22: You need to define return type, even if it's void, to make things easier to follow and to have the compiler properly understand what you're doing.

apps/client/src/app/modules/list/list-details-panel/list-details-panel.component.ts
```diff
@@ -455,6 +486,31 @@ export class ListDetailsPanelComponent implements OnChanges, OnInit {
     this.message.success(this.translate.instant('LIST.Copied_as_text'));
   }
 
+  // Done to reduce render-time perf of generating the whole string into the DOM. If there's a way to do it with [cdk directive](https://material.angular.io/cdk/clipboard/overview) that would be ideal
+  public copyJSONExport(serverName: string) {
```
Supamiu, line 490:
```suggestion
  public copyJSONExport(serverName: string): void{
```

apps/client/src/app/pages/list-details/list-details/list-details.component.html
```diff
@@ -184,14 +184,24 @@ <h2>{{list.name}}</h2>
...
+            <!-- cannot have async pipes in a template action https://stackoverflow.com/questions/40300115/how-to-pass-async-variable-in-template-action-function -->
+            <input  #servername [value]="server$ | async" type="hidden" />
+            <button (click)="$event.stopPropagation();copyJSONExport(servername.value, listDisplay, list)"
+                    *ngIf="display$ | async as listDisplay"
+                    [nzTitle]="'LIST.Copy_list_as_json' | translate"
+                    nz-button nz-tooltip
+                    nzShape="circle"
+                    nzSize="small">
+              <i nz-icon nzType="snippets"></i>
+            </button>
```
Supamiu, line 204: You can simply wrap this inside a parend container component, having hidden components is too hacky for me to accept it ;)

```suggestion
          <ng-container *ngIf="server$ | async as server">
            <button (click)="$event.stopPropagation();copyJSONExport(server, listDisplay, list)"
                    *ngIf="display$ | async as listDisplay"
                    [nzTitle]="'LIST.Copy_list_as_json' | translate"
                    nz-button nz-tooltip
                    nzShape="circle"
                    nzSize="small">
              <i nz-icon nzType="snippets"></i>
            </button>
          </ng-container>
```

apps/client/src/app/pages/list-details/list-details/list-details.component.ts
```diff
@@ -316,6 +335,11 @@ export class ListDetailsComponent extends TeamcraftPageComponent implements OnIn
     return (remaining > 0) ? (exportString + `${remaining}x ${this.i18nTools.getName(this.l12n.getItem(row.id))}\n`) : exportString;
   }
 
+  public copyTextExport(display: ListDisplay, list: List) {
```
Supamiu, line 338:
```suggestion
  public copyTextExport(display: ListDisplay, list: List): void {
```

apps/client/src/app/pages/list-details/list-details/list-details.component.ts
```diff
@@ -332,6 +356,26 @@ export class ListDetailsComponent extends TeamcraftPageComponent implements OnIn
     this.message.success(this.translate.instant('LIST.Copied_as_text'));
   }
 
+  public copyJSONExport(serverName: string, display: ListDisplay, list: List) {
```
Supamiu, line 359:
```suggestion
  public copyJSONExport(serverName: string, display: ListDisplay, list: List) : void {
```

apps/client/src/app/pipes/pipes/mob-name.pipe.ts
```diff
@@ -1,18 +1,22 @@
 import { Pipe, PipeTransform } from '@angular/core';
+
 import { LocalizedDataService } from '../../core/data/localized-data.service';
 
 @Pipe({
   name: 'mobName'
 })
 export class MobNamePipe implements PipeTransform {
-
+  public static GetActualMobId(id: number) {
```
Supamiu, line 9: Why make this `public`? I can understand `static` for small optimization but not `public`. Also, in Typescript, method names are lowerCamelCase, not UpperCamelCase.

apps/client/src/app/pages/list-details/list-details/list-details.component.ts
```diff
@@ -332,6 +356,26 @@ export class ListDetailsComponent extends TeamcraftPageComponent implements OnIn
...
+    if (this._clipboardService.copyFromContent(this.getListJsonExport(serverName, display, list)))
+      this.afterListJSONCopied();
+  }
+
+  public getListJsonExport(serverName: string, display: ListDisplay, list: List): string {
+    const seed = list.items.filter(row => row.id < 20).reduce((exportString, row) => {
+      return this.appendExportStringWithRow(exportString, row);
+    }, `${this.translate.instant('Crystals')} :\n`) + '\n';
+    return JSON.stringify(this.getSerializedRowData(serverName, list.items, list.finalItems))//,null,2);
```
Supamiu, line 368:
```suggestion
    return JSON.stringify(this.getSerializedRowData(serverName, list.items, list.finalItems));
```

Supamiu (changes requested):

apps/client/src/app/modules/list/data/ListRowSerializationHelper.ts
iturtleman: currently it accepts anything that has an id or idName and returns whatever the input is so should I make it semi explit or just put `any`

apps/client/src/app/pipes/pipes/mob-name.pipe.ts
iturtleman: I used it in `E:\ffxiv-teamcraft\apps\client\src\app\modules\list\data\ListRowSerializationHelper.ts`
and my bad on naming. I'm using primarily c# so I forget when I context switch.

apps/client/src/app/modules/list/data/ListRowSerializationHelper.ts
```diff
@@ -0,0 +1,315 @@
...
+  private getMobNameFromId(id: number): I18nName {
+    return this.l12n.getMob(MobNamePipe.getActualMobId(id));
+  }
+
+  private getMonsterHuntData(monsterDrops: MonsterInfo[]): MonsterInfoWithMobName[] {
+    return monsterDrops && monsterDrops.length > 0 ? monsterDrops.map((r: any) => this.applyMonsterInfo(r)) : undefined;
+  }
+
+  public serializeNPCs(vendors: Vendor[]): VendorWithName[] {
+    return vendors && vendors.length > 0 ? vendors.map((r: Vendor) => this.applyNpcName(r)) : undefined;
```
Supamiu, line 136: I'm not a big fan of returning undefined in the case where the array is empty, this would transform `[]` into `undefined`, breaking any call made using the Array prototype.

apps/client/src/app/modules/list/data/ListRowSerializationHelper.ts
```diff
@@ -0,0 +1,315 @@
...
+  }
+
+  private serializeVentures(item: ListRow, ventures: number[]) {
```
Supamiu, line 139: Please add a return type.

apps/client/src/app/modules/list/data/ListRowSerializationHelper.ts
```diff
@@ -0,0 +1,315 @@
...
+  private getNameIfExists(name: I18nName): string {
+    const translatedName = this.i18nTools.getName(name);
+    return translatedName && translatedName !== 'no name' ? translatedName : undefined;
+  }
+
+  private serializeVoyages(voyages: any): string {
+    return voyages && voyages.length > 0 ? voyages.map((r: any) => this.getNameIfExists(r)) : undefined;
+  }
+
+  public applyNpcName(obj: Vendor): any {
```
Supamiu, line 80: You have plenty of interfaces at the start of the file, I would expect return types not to be any is such a context.

apps/client/src/app/modules/list/data/ListRowSerializationHelper.ts
```diff
@@ -0,0 +1,315 @@
...
+    return {
+      ...obj,
+      npcName: this.getNameIfExists(this.l12n.getNpc(obj.npcId)),
+      zoneName: this.getZoneName(obj),
+    };
+  }
+
+  private getZoneName(obj: MapInfo): string {
+    if (!obj)
+      return undefined;
```
Supamiu, line 90: Return type says string but you're returning undefined here, this means that I should be able to do `getZoneName().toLowercase()` for instance, based on the typing it wouldn't crash, but it can because in reality, the return type doesn't match what can really be returned.

apps/client/src/app/modules/list/data/ListRowSerializationHelper.ts
```diff
@@ -0,0 +1,315 @@
...
+
+  public serializeNPCs(vendors: Vendor[]): VendorWithName[] {
+    return vendors && vendors.length > 0 ? vendors.map((r: Vendor) => this.applyNpcName(r)) : undefined;
+  }
+
+  private serializeVentures(item: ListRow, ventures: number[]) {
+    return ventures && ventures.length > 0 ? this.gt.getVentures(ventures).map(venture => {
+      const retval = {
+        ...venture,
+        ventureDetails: VenturesComponent.ventureAmounts(venture)
```
Supamiu, line 143: Calling a static method from a component here is not acceptable from a design pattern point of view. Angular provides dependency injection and Typescript provides inheritence so you can avoid referencing static code accross multiple classes, making dependncies way easier to understand and organize.

apps/client/src/app/modules/list/data/ListRowSerializationHelper.ts
```diff
@@ -0,0 +1,315 @@
...
+
+  private applyIntanceName(instance: Instance): Instance {
+    const retval = { ...instance }
+    if (instance != null && !instance.name)
+      retval.name = this.getNameIfExists(this.l12n.getInstanceName(instance.id))
+    return retval;
+  }
+
+  private serializeInstances(instances: Instance[]): InstanceWithName[] {
+    return instances && instances.length > 0 ? instances.map((r: Instance) => this.applyIntanceName(r)) : undefined;
```
Supamiu, line 169: Same remark as the monsters serializer, according to the signature, undefined isn't a possible return value, yet it is according to the implementation. this is the same for all the methods below but I don't copy this comment on every single one of them.

apps/client/src/app/modules/list/data/ListRowSerializationHelper.ts
```diff
@@ -0,0 +1,315 @@
...
+
+  private serializeTreasures(treasures: ItemNameHolder[]): ItemNameHolder[] {
+    return treasures && treasures.length > 0 ? treasures.map((r: any) => this.applyItemName(r)) : undefined;
+  }
+
+  private serializeMasterbooks(masterbooks: ItemNameHolder[]): ItemNameHolder[] {
+    return masterbooks && masterbooks.length > 0 ? masterbooks.map((r: any) => this.applyItemName(r)) : undefined;
+  }
+
+  private serializeGardening(gardening: any) {
```
Supamiu, line 184: Please add proper return type.

apps/client/src/app/modules/list/data/ListRowSerializationHelper.ts
```diff
@@ -0,0 +1,315 @@
...
+    if (!Number.isInteger(gardening))
```
Supamiu, line 185: If you don't accept anything else than number, type the input to `number` instead of `any` plus this code logic.

apps/client/src/app/modules/list/data/ListRowSerializationHelper.ts
```diff
@@ -0,0 +1,315 @@
...
+      marketBoardLink: `https://universalis.app/market/${item.id}`,
+    };
+    //get rid of fields that are confusing for an export layer
+    delete retval.sources;
+    delete retval.craftedBy;
+    delete retval.amount_needed;
+    return retval;
+  }
+
+  private serializeAchievements(achievements: any) {
```
Supamiu, line 295: Please add proper typing, this applies to all the methods below.

apps/client/src/app/modules/list/data/ListRowSerializationHelper.ts
```diff
@@ -0,0 +1,315 @@
...
+      .map((item: ListRow) => {
+        const desynths = ListRowSerializationHelper.getData(item, DataType.DESYNTHS);
+        return desynths && desynths.length > 0 ? desynths.map((r: any) => this.applyItemName({ id: r })) : undefined
+      })
+      .filter(m => !!m)
+      .reduce((pn, u) => [...pn, ...u], []);
+    for (let i = 0; i < desynthData.length; i++) {
+      desynthMap[desynthData[i].id] = desynthData[i].name;
+    }
+    const retval = {
```
Supamiu, line 230: I don't understand the purpose of `retval = ...; return retval;`, why not return the value directly?

apps/client/src/app/modules/list/data/ListRowSerializationHelper.ts
```diff
@@ -0,0 +1,315 @@
...
+interface MonsterInfoWithMobName extends MonsterInfo, NameHolder { }
+interface VendorWithName extends Vendor, NameHolder { }
+interface InstanceWithName extends Instance, NameHolder { }
+interface GatherWithName extends GatheredBy, NameHolder { }
+
+export class ListRowSerializationHelper {
+
+  public static getData<T = any>(row: ListRow, type: DataType, isObject = false): T {
+    return getItemSource<T>(row, type, isObject);
+  }
```
Supamiu, line 33: What's the purpose of this method? You can also just use `getItemSource` directly.

apps/client/src/app/modules/list/list-details-panel/list-details-panel.component.html
```diff
@@ -101,12 +101,18 @@
                   nzSize="small">
             <i class="material-icons">local_atm</i>
           </button>
-          <button (cbOnSuccess)="textCopied()" (click)="$event.stopPropagation()" [cbContent]="getTextExport()"
+          <button (click)="$event.stopPropagation();copyTextExport()"
```
Supamiu, line 104: So I thought a bit more about this clipboard issue and I think we should have our own lazy clipboard directive, instead of having to import the service in every single component that does clipboard copy stuff.

apps/client/src/app/modules/list/list-details-panel/list-details-panel.component.ts
```diff
@@ -95,13 +103,22 @@ export class ListDetailsPanelComponent implements OnChanges, OnInit {
 
   hasAlreadyBeenOpened: boolean;
 
+  server$ = this.authFacade.mainCharacter$.pipe(
+    map(char => char.Server)
+  )
+  private serializationHelper = new ListRowSerializationHelper(this.i18nTools, this.l12n, this.gt);
```
Supamiu, line 109: Why are you using the helper like this instead of making it a service and let Angular handle dependencies for you (see any `.service.ts` file for example use of the dependency injection)?

apps/client/src/app/modules/list/list-details-panel/list-details-panel.component.ts
```diff
@@ -180,22 +194,34 @@ export class ListDetailsPanelComponent implements OnChanges, OnInit {
       this.hasNavigationMap = this.getZoneBreakdownPathRows(this.zoneBreakdown).length > 0;
     }
     this.hasTrades = this.displayRow.rows.reduce((hasTrades, row) => {
-      return (this.getData(row, DataType.TRADE_SOURCES).length > 0) || (this.getData(row, DataType.VENDORS).length > 0) || hasTrades;
+      return ListRowSerializationHelper.getData(row, DataType.TRADE_SOURCES).length > 0
```
Supamiu, line 197: Why are you changing these calls to a static method in the serialization helper that just calls the getDataSource method directly?

apps/client/src/app/pages/list-details/list-details/list-details.component.ts
```diff
@@ -98,16 +108,25 @@ export class ListDetailsComponent extends TeamcraftPageComponent implements OnIn
 
   private regeneratingList = false;
 
+  server$ = this.authFacade.mainCharacter$.pipe(
+    map(char => char.Server)
+  )
+  private serializationHelper = new ListRowSerializationHelper(this.i18nTools, this.l12n, this.gt);
```
Supamiu, line 114: Same as for the list-details component, use DI instead of doing this please.

apps/client/src/app/pages/list-details/list-details/list-details.component.ts
```diff
@@ -317,6 +336,11 @@ export class ListDetailsComponent extends TeamcraftPageComponent implements OnIn
     return (remaining > 0) ? (exportString + `${remaining}x ${this.i18nTools.getName(this.l12n.getItem(row.id))}\n`) : exportString;
   }
 
+  public copyTextExport(display: ListDisplay, list: List): void {
+    if (this._clipboardService.copyFromContent(this.getListTextExport(display, list)))
```
Supamiu, line 340: Even if there's only one instruction, I'd prefer braces over here.

apps/client/src/app/pages/list-details/list-details/list-details.component.ts
```diff
@@ -333,6 +357,26 @@ export class ListDetailsComponent extends TeamcraftPageComponent implements OnIn
     this.message.success(this.translate.instant('LIST.Copied_as_text'));
   }
 
+  public copyJSONExport(serverName: string, display: ListDisplay, list: List) : void {
+    if (this._clipboardService.copyFromContent(this.getListJsonExport(serverName, display, list)))
```
Supamiu, line 361: Even if there's only one instruction, I'd prefer braces over here.

apps/client/src/app/pipes/pipes/mob-name.pipe.ts
```diff
@@ -1,17 +1,23 @@
...
 
 @Pipe({
-  name: 'mobName',
+  name: 'mobName'
 })
 export class MobNamePipe implements PipeTransform {
-  constructor(private data: LocalizedLazyDataService) {}
-
-  transform(id: number): I18nNameLazy {
+  public static getActualMobId(id: number) {
```
Supamiu, line 9: Why a static method here? I could see this inside the LazyDataService but static is soemthing to avoid in Angular, because Dependency injection is here to handle code that can be reused in different places, while static context cannot be mocked and makes it hard to test and follow the implementations and the dependency graph.

Supamiu (changes requested):

apps/client/src/app/modules/list/data/ListRowSerializationHelper.ts
iturtleman: good catch missed that once I figured out the correct type.

apps/client/src/app/modules/list/data/ListRowSerializationHelper.ts
iturtleman: I will solidify the rest of these. I was looking for existing types where possible but I struggled to find some.

apps/client/src/app/modules/list/data/ListRowSerializationHelper.ts
iturtleman: ah! so I should explicitly put `string|undefined` the intent is to not generate the key in the final JSON

apps/client/src/app/modules/list/list-details-panel/list-details-panel.component.ts
iturtleman: not realizing this could be entirely replaced with `getItemSource<Type>` and since I had moved `getData` I had the old code reference the new location. this should get reverted or call `getItemSource` directly. not sure why it wasn't from the getgo.

apps/client/src/app/pipes/pipes/mob-name.pipe.ts
iturtleman: lack of understanding about statics in angular

apps/client/src/app/modules/list/list-details-panel/list-details-panel.component.ts
Supamiu: Using `getItemSource` is the best way to do it, I used to use a custom method inside the component when I first implemented this logic, then I changed it but didn't migrate this part.

Supamiu:
With #1645 , you'll be able to use the lazy directive directly instead of having to call the service for performance improvements 🙂

Supamiu:
I'm closing this PR for cleanup purpose and inactivity, also not sure it'd be integrated as such inside the app, as discussed in the past on Discord.

PR #1625, opened 2020-09-11
Fishing spot page data loading refactoring, lazy data

emily-curry:
Sorry for another huge PR! I had intended to keep this down to just the fishing spot page, but found a few bugs relating to `I18nName` values not updating when languages were switched (requiring a page refresh), so I decided to move forward refactoring all the pipes to return `I18nNameLazy` instead, which lets their values change as the language changes. This turned out to be a pretty safe refactor, as in nearly all cases, the only thing being done with the result of pipes that returned `I18nName` was piping that value to `i18n` (which now handles those lazy values). I also noticed a few places where observables are likely leaking (imperative `.subscribe()` calls without use of `takeUntil` or `Subscription.unsubscribe`), and there are probably more out there. It's something I'll be on the lookout for.


emily-curry:
image
oh my goodness I'm so sorry 😭

apps/client/src/app/core/data/localized-lazy-data-service.spec.ts
```diff
@@ -17,7 +19,7 @@ const languages = ['en', 'fr', 'ja', 'de', 'ko', 'zh'] as const; // Not testing
  * Note that this must remain separate from the other load function, as the path names must be statically analyzable for webpack to include the resources in the bundle.
  * @param key The lazy data key to load data for.
  */
-const load = async (key: LazyDataKey) => {
+const loadStd = async (key: LazyDataKey) => {
   return await import('../../../assets/data/' + lazyFilesList[key].fileName);
 };
```
Supamiu, line 24: This is going to be a big issue, because it defeats the purpose of lazy files. I originally created this system because large jsons were imported as-is and they were making the build take way too much RAM, which broke the build process inside circleci.

Lazy Data aims at providing a lighter bundle and faster build times by excluding json files from the build process and loading them at runtime from the assets instead.

This change it probably the reason why it breaks inside the CI as of now.

apps/client/src/app/pages/db/fishing-spot/fishing-spot-bait-datagrid/fishing-spot-bait-datagrid.component.ts
```diff
@@ -0,0 +1,45 @@
...
+    <ng-template #iconRender let-id>
+      <app-item-icon [itemId]="id" width="32"></app-item-icon>
+    </ng-template>
+    <app-fishing-spot-datagrid
+      [activeFish]="activeFish"
+      (activeFishChange)="activeFishChange.next($event)"
+      [rowIconRender]="iconRender"
+      [table]="table$ | async"
+    ></app-fishing-spot-datagrid>
+  </nz-card>`,
```
Supamiu, line 24: Is there a specific reason for this template to be inline rather than inside its own file?

apps/client/src/app/core/data/localized-lazy-data-service.spec.ts
emily-curry: Does this apply to the unit test files as well? I can't say I know enough about how the bundler works in the test runner to know for sure. 
The motivation for this was that I didn't want to rely on the same mechanisms the app relied on to load the data, as it might increase the possibility of false negatives.

apps/client/src/app/pages/db/fishing-spot/fishing-spot-bait-datagrid/fishing-spot-bait-datagrid.component.ts
emily-curry: Nope! I'll break it out into it's own file.

apps/client/src/app/core/data/localized-lazy-data-service.spec.ts
Supamiu: I'm not sure neither, but your circleci build fails because of an oom error, so I guess it is possible somehow oO

emily-curry: Interesting! The test stage passed, but the build itself failed (and these .spec.ts files shouldn't be included in the build). I'll look into it.

emily-curry:
Marking this as ready for review. It's code-complete, but I am still looking into the build issue. If I don't turn up anything soon, I'll just use the lazy data provider service to load the data in tests and see if that fixes it.

apps/client/src/app/core/data/localized-lazy-data-service.spec.ts
emily-curry: The build did not time out this time, for no discernable reason. 💁‍♀️

apps/client/src/app/pipes/pipes/quest-icon.pipe.ts
```diff
@@ -1,17 +1,38 @@
...
+import { distinctUntilChanged, map, switchMap } from 'rxjs/operators';
+import { LocalizedLazyDataService } from '../../core/data/localized-lazy-data.service';
 
 @Pipe({
   name: 'questIcon',
-  pure: true
+  pure: false,
 })
-export class QuestIconPipe implements PipeTransform {
+export class QuestIconPipe implements PipeTransform, OnDestroy {
```
Supamiu, line 10: I feel like some pipes have common code for everything related to subscriptions and all that, maybe creating a super class for them could be handy for future updates.

emily-curry: Hm, I'll think on this one! It can be a little awkward for the same reason inheritance is awkward in components, when you override lifecycle methods it can be pretty bad if you forget to call the superclass', but there's probably a composition-based approach that shouldn't have that issue.

emily-curry:
@Supamiu I went over this once more tonight, going through most of the flows in the app. Fixed a few small bugs, but at this point I am feeling good about it.

apps/client/src/app/pages/db/db-comments/db-comments.service.ts
```diff
@@ -8,29 +8,34 @@ import { Observable } from 'rxjs';
...
       .snapshotChanges()
       .pipe(
         map((snaps: DocumentChangeAction<DbComment>[]) => {
-          const comments = snaps
-            .map((snap: DocumentChangeAction<any>) => {
-              const valueWithKey: DbComment = <DbComment>{ ...snap.payload.doc.data(), $key: snap.payload.doc.id };
-              delete snap.payload;
-              return valueWithKey;
-            });
+          console.log(snaps);
```
Supamiu, line 29: oops

apps/client/src/app/pages/db/db-comments/db-comments.service.ts
```diff
@@ -8,29 +8,34 @@ import { Observable } from 'rxjs';
...
+          const comments = snaps.map((snap: DocumentChangeAction<any>) => {
+            const valueWithKey: DbComment = <DbComment>{ ...snap.payload.doc.data(), $key: snap.payload.doc.id };
+            delete snap.payload;
+            return valueWithKey;
+          });
           return this.serializer.deserialize<DbComment>(comments, [this.getClass()]);
         }),
-        map(comments => {
+        map((comments) => {
+          console.log(comments);
```
Supamiu, line 38: oops

apps/client/src/app/pages/db/db-comments/db-comments/db-comments.component.ts
```diff
@@ -64,37 +59,27 @@ export class DbCommentsComponent extends TeamcraftComponent implements OnInit {
...
-
-    this.translate.onLangChange.pipe(
-      takeUntil(this.onDestroy$)
-    ).subscribe((change) => {
-      lang$.next(change.lang);
-    });
-
-    const comments$ = combineLatest([contentComments, lang$]).pipe(
       map(([comments, lang]) => {
+        console.log(comments, lang);
```
Supamiu, line 82: Seems like there's a lot of remaining logs 😄

Supamiu (changes requested):

apps/client/src/app/pages/db/db-comments/db-comments/db-comments.component.ts
emily-curry: 🤦‍♀️ resolved.

Supamiu (approved):

PR #1632, opened 2020-09-15
Draft: Auto-generate lazy data types proof-of-concept

emily-curry:
Opening to gather feedback on the approach. How do you feel about the naming scheme? I'd like to discourage direct use of these types, and instead rely on services such as the LocalizedLazyDataService to convert data from our generated types to our domain types.

Build failures are to be expected at this point, this causes a lot of squiggles.


apps/client/src/app/core/data/lazy-data.service.ts
```diff
@@ -146,27 +149,36 @@ export class LazyDataService {
       }),
       filter((recipes) => recipes !== undefined),
       map((recipes) => {
-        return recipes.find((r) => r.id.toString() === id.toString()) || this.data.recipes.find((r) => r.id.toString() === id.toString());
+        return (
+          ((recipes as unknown) as Craft[]).find((r) => r.id.toString() === id.toString()) ||
+          ((this.data.recipes as unknown) as Craft[]).find((r) => r.id.toString() === id.toString())
```
Supamiu, line 154: Why do we have to cast as unknown before casting as `Craft[]`? The recipe model and the craft one should match, if they don't, the Craft model needs to be readjusted because it's used by simulator where we feed recipes ^^

Supamiu (changes requested):

apps/client/src/app/core/data/lazy-data.service.ts
emily-curry: Yep, you are correct, they don't match up quite right. Craft actually comes from the sim, so I can't fix it here.

But that makes these auto-generated types really valuable for validating our assumptions about the shape of this json data. Something we could do to fix it here is to just write an adapter from lazyrecipe->craft, that way we're not breaking everything that already relies on craft.

Supamiu:
I really like the idea of having generated types from game data, do you want to push this further into a production-ready solution?

I see tests are failing but this can be fixed pretty easily, also the craft type issue can be fixed by either changing it inside simulator's types or by writing an adapter as you suggested 😄

Supamiu:
Haha yeah I created a bunch of new issues for hacktoberfest, feel free to ping me on discord if you want some stuff to do 👍

Supamiu:
I'm closing this for cleanup purpose, will reopen if it turns into a real thing 👍

PR #1647, opened 2020-10-02
Fix/smart quotes in macro translator

PaddyS:
The macro translator fails (silently, just a console error) when using smart quotes, such as:
```
// Smart quotes
/ac “Inner quiet” <me><wait.2>
// Normal quotes for comparison
/ac "Inner quiet" <me><wait.2>
```

This PR adds support for those smart quotes, as well as an e2e test for that case.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,12 +10,12 @@ Simply run `npm start` to start a live server with file watcher.
 
 ### Branch names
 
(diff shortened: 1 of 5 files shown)


Supamiu (approved):

Supamiu:
That's an impressive PR with nice e2e contribution ! 

Thank you for that, if you want to work more on e2e testing you'd be welcome for sure 👍

PR #1799, opened 2021-01-19
Feat/submarine exploration result reporter

shambarick:
Add a reporter for submarine voyage log to track loots.


apps/client/src/app/core/data-reporting/exploration-result.reporter.ts
```diff
@@ -0,0 +1,8 @@
+export enum ExplorationType {
+  AIRSHIP,
+  SUBMARINE
+}
```
Supamiu, line 4: Even if it may seem a bit too much, one file per export is a general rule i'm trying to follow (well, sometimes I don't, but it's rare), I'd love this to be in its own file.

apps/client/src/app/core/data-reporting/submarine-exploration-result-reporter.ts
```diff
@@ -0,0 +1,131 @@
...
+import { Observable } from 'rxjs/Observable';
+import { ofPacketType } from '../rxjs/of-packet-type';
+import { filter, map, shareReplay, startWith, tap, withLatestFrom } from 'rxjs/operators';
+import { merge } from 'rxjs';
+import { LazyDataService } from '../data/lazy-data.service';
+import { ExplorationResultReporter, ExplorationType } from './exploration-result.reporter';
+
+export class SubmarineExplorationResultReporter implements DataReporter, ExplorationResultReporter {
+  private statusList$;
+  private resultLog$;
```
Supamiu, line 11: Could be nice having proper typings for these, they're implicit any and it makes the code harder to read and maintain.

apps/data-extraction/src/extractors/submarine-ranks.extractor.ts
```diff
@@ -0,0 +1,29 @@
...
+  protected doExtract(): any {
+    const ranks = {};
+
+    this.getAllPages(this.getResourceEndpointWithQuery(XivapiEndpoint.SubmarineRank, {
+      columns: 'ID,Capacity,ExpToNext,SurveillanceBonus,RetrievalBonus,SpeedBonus,RangeBonus,FavorBonus',
+    })).pipe(
+      map(((page) => page.Results)),
+      tap((rankResults)  => {
+        rankResults.forEach((result) => {
+          ranks[result.ID] = result;
```
Supamiu, line 16: Whenever it's possible, prefer a custom structure so if tomorrow things are renamed for whatever reason, only this part of the code will have to change. This also helps with naming convention as TC uses lowercase for property names while XIVAPI uses UpperCamelCase.

apps/data-extraction/src/extractors/submarine-parts.extractor.ts
```diff
@@ -0,0 +1,50 @@
...
+          ids: itemResults.map((r) => r.AdditionalData).join(','),
+          columns: 'ID,Components,Favor,Range,Rank,RepairMaterials,Retrieval,Slot,Speed,Surveillance'
+        }))
+          .pipe(
+            map((page) => page.Results),
+            tap((partResults) => {
+              partResults.forEach((part) => {
+                parts[part.ID] = {
+                  ...part,
+                  ItemID: itemResults.filter((r) => r.AdditionalData === part.ID)[0]['ItemID']
```
Supamiu, line 34: Having lazy files follow the lowercase naming convention would be nice, also same comment as for the other exporter regarding custom structures instead ^^.

apps/data-extraction/src/extractors/submarine-ranks.extractor.ts
shambarick: Np. Isn't there something that can be used for checking the struct so the extractor throws an error if the response doesn't contain the columns? Like an endpoint in XIVAPI that returns the body spec or just the list of the fields for a given index (eg: SubmarineRanks)
That would avoid the trouble of checking if there is struct changes manually every update.

apps/data-extraction/src/extractors/submarine-parts.extractor.ts
shambarick: Yep, I noticed that a bit too late and I already pushed the code, was gonna fix that for the next push.

apps/data-extraction/src/extractors/submarine-ranks.extractor.ts
Supamiu: I'm the maintainer of XIVAPI so I'll be aware of changes since I'm the one making them 😉 to get the fields for a given sheet, just get one of the items, like https://xivapi.com/SubmarineRank/1, there's no optional fields, so every single field can be found on every single entry 👍

apps/client/src/app/core/data-reporting/exploration-result.reporter.ts
shambarick: For now, I have put it inside `model/other`. Might actually be better to move it in a `airship/submarine` module tho.

shambarick:
@Supamiu , should be good now

apps/data-extraction/src/extractors/submarine-parts.extractor.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  protected doExtract(): any {
+    const parts = {};
+
+    this.getAllPages(this.getSearchEndpointWithQuery({
+      indexes: XivapiEndpoint.Item,
+      columns: 'ID,AdditionalData',
+      string_column: 'FilterGroup',
+      string: '36',
+      string_algo: 'match'
```
Supamiu, line 15: You should use filters instead, this is the query string fragment for it: `filters=FilterGroup=36`, matching a number using a fulltext search is not the way to go for this one :)

apps/data-extraction/src/main.ts
```diff
@@ -123,7 +127,7 @@ const operationsSelection = new MultiSelect({
     'everything',
     ...extractors.map(extractor => {
       return extractor.getName();
-    })
+    }).sort((a, b) => a.localeCompare(b)),
```
Supamiu, line 130: The extractors are listed in order to avoid having issues with extractor dependencies. The mappy and LGB extractors require informations from previous extractors, which is why the array isn't sorted programatically, it's made by hand.

Supamiu (changes requested):
I see you implemented a reporter but for submarine only, are you planning the airship one for later on or should I wait for it to merge?

apps/data-extraction/src/main.ts
shambarick: Sorry, I didn't know :/

shambarick:
You can merge as is.

The airship one is for later because I still need a few more samples for the struct I'm missing. So I have no ETA for that one atm.

Supamiu (approved):

PR #1855, opened 2021-03-11
Styling changes

ImCheesecake:
Fixed some styling issues on mobile/tablet. Made the side-menu a bit wider on desktop and added z-index on mobile.


apps/client/src/app/app.component.html
```diff
@@ -202,6 +202,7 @@
                 *ngIf="!overlay"
                 [(nzCollapsed)]="settings.compactSidebar"
                 [nzCollapsedWidth]="80 | ifMobile: 0"
+                [nzWidth]="255"
```
Supamiu, line 205:
```suggestion
                [nzWidth]="255 | ifMobile: 200"
```

I think it's better to keep it 200 on mobile, the ifMobile pipe being good for this kind of use case :)

apps/client/src/theme.less
```diff
@@ -336,6 +336,7 @@ a, .ant-table-filter-dropdown-link {
 
 .ant-layout-sider {
   background: var(--topbar-color);
+
```
Supamiu, line 339:
```suggestion
```

Supamiu (changes requested):
First static review without running it locally ^^

apps/client/src/app/app.component.html
ImCheesecake: Fair enough, I had that before but since I added the z-index I thought it could be wider :)

Supamiu (approved):

PR #1856, opened 2021-03-12
style(layouts): Refactoring

ImCheesecake:
I've noticed that some UI elements doesn't work really well between desktop and handheld devices so I've decided to make some refactoring that might fix them. Most changes shouldn't impact the users at all, but there might be some components that is being moved around to make it more responsive between devices.

Below is the changes that I've currently made. 
# WIP

### After my rework

<hr>


Supamiu:
Closing for inactivity for now, seems like the progress on this is stopped for now, will be happy to reopen if you decide to push this further to make it ready for production 👍

PR #1883, opened 2021-04-10
feat(packet-capture): allow loading local opcodes for testing

zhyupe:
* This PR requires **pcap-ffxiv** being published and updated.
* I have no idea how to build electron app at local, so this is not fully tested. As a result this change may require extra testing.

Passes `localOpcodesPath: join(process.resourcesPath, 'localOpcodes')` to **pcap-ffxiv**. Therefore opcodes can be tested before published by putting the files in `%tc-root%/app-%version%/resources/localOpcodes`.

### DIFF
--- a/apps/electron/src/pcap/packet-capture.ts
+++ b/apps/electron/src/pcap/packet-capture.ts
@@ -1,7 +1,7 @@
 import { exec } from 'child_process';
 import { MainWindow } from '../window/main-window';
 import { Store } from '../store';
-import { join } from 'path';
+import { join, resolve } from 'path';
 import { app } from 'electron';
 import * as isDev from 'electron-is-dev';
 import * as log from 'electron-log';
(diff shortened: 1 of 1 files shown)


apps/electron/src/pcap/packet-capture.ts
```diff
@@ -157,7 +158,11 @@ export class PacketCapture {
       options.pid = this.options.pid;
     }
 
-    this.captureInterface = new CaptureInterface(isDev ? options : { ...options, exePath: PacketCapture.MACHINA_EXE_PATH });
+    this.captureInterface = new CaptureInterface(isDev ? options : { 
+      ...options, 
+      exePath: PacketCapture.MACHINA_EXE_PATH,
+      localOpcodesPath: PacketCapture.LOCAL_OPCODES_PATH,
+    });
```
Supamiu, line 165: You're making it use local opcodes if the env is not dev, which means local opcodes in production. That's not the way it's designed, loading from network has to stay the default behavior, also I'm not sure forcing local opcodes in dev env is the way to go, not everyone starts local dev env to test opcodes ^^

Supamiu (changes requested):

apps/electron/src/pcap/packet-capture.ts
zhyupe: The local opcode was designed to be placed manually, which means the behaviour is enforced, not acting as a fallback, and only developers who want to test the opcode should touch the file. If the file not exists, it would fallback to network requests.

But I do agree that it should be placed in dev and have an environment variable which controls this behaviour. Would work on this tomorrow.

Supamiu:
Regarding how to test the app on local env:

 - Run `build:watch` script, wait for it to finish the intial build.
 - Run `electron:start`.
 
 This will start a new electron app with its context (the place where electron stuff and cache files are) in `%AppData%/electron`.

Supamiu (approved):

PR #1896, opened 2021-04-18
refactor(submarine-reporter): better observable chain for the reporter and unicity for reports

Supamiu:
@shambarick You'll want to check this one 🙂

### DIFF
--- a/apps/client/src/app/core/data-reporting/airship-exploration-result-reporter.ts
+++ b/apps/client/src/app/core/data-reporting/airship-exploration-result-reporter.ts
@@ -37,16 +37,6 @@ export class AirshipExplorationResultReporter extends ExplorationResultReporter 
       toIpcData()
     );
 
-    // Undefined usage for now, will keep in here just in case
-    // const updateHullCondition$ = packets$.pipe(
-    //   ofMessageType('updateInventorySlot'),
-    //   toIpcData(),
-    //   withLatestFrom(isAirshipMenuOpen$),
-    //   filter(([updateInventory, isOpen]) => {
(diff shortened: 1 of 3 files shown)


apps/client/src/app/core/data-reporting/airship-exploration-result-reporter.ts
```diff
@@ -37,23 +38,20 @@ export class AirshipExplorationResultReporter extends ExplorationResultReporter
...
-        return this.getBuildStats(status.hull, status.rigging, status.forecastle, status.aftcastle);
-      }),
-      switchMap(stats => {
-        return resultLog$.pipe(
-          map(resultLog => this.createReportsList(stats, resultLog))
+      filter(([, isOpen]) => isOpen),
+      switchMap(([resultLog]) => {
+        return status$.pipe(
+          filter((status: AirshipStatus) => {
+            return this.shouldSendReport(status.returnTime, resultLog.airshipSpeed, resultLog.explorationResult);
```
shambarick, line 47: Aaah.... How fun, the `AirshipStatus` doesn't have the `birthDate`. It might be better to do like submarines then. Use the `UpdateInventorySlot` to figure out the ship slot then get its `birthDate` from the `AirshipStatusList`

shambarick (changes requested):

apps/client/src/app/core/data-reporting/airship-exploration-result-reporter.ts
Supamiu: I think the speed is enough, because returnTime + result is already almost good enough for Unicity imo.

PR #1900, opened 2021-04-23
fix(simulator): Prevent tooltip leak

infinitestory:
Currently, drag and drop can cause the mouseleave event to be missing if the action element moves. This leaks the tooltip overlay element when mousing over the action again. Prevent this by checking for an existing OverlayRef when creating a tooltip.


Supamiu (approved):
Thanks for this nice fix, will me in the next update

PR #1980, opened 2021-08-31
fix(lists): item craft ready blue highlight fix

### LINKED ISSUE #1721: bug: Blue "Materials Acquired" Indicators do not Account for Consumed Materials
**Describe the bug**
When working to craft items in a list, there are two visual indicators to say "you have the materials to craft this item": a dotted blue line around the item if precrafts are needed before making it, and a solid blue background if the craft can be started immediately.  These indicators do not account for if those materials were consumed in crafting a different item, and therefore no longer available.

**To Reproduce**
1. Create a new list.
2. In this list, add 1 Darksteel Rings and 1 Darksteel Rivets as final products.
3. The list will populate with 2 Darksteel Nuggets as required precrafts, 1 for each final product.
4. Add as external exactly 1 Darksteel Nugget to the required precrafts, enough to complete either but not both final products.  Observe that both final products (correctly) light up with solid blue backgrounds.
5. Check off the Darksteel Rings, consuming the Darksteel Nugget to complete the item.  Open the Inventory view of the list and verify that no Darksteel Nuggets are still present to be crafted with.
6. Observe that the Darksteel Rivets still have a solid blue background, even though the list has been told that all acquired nuggets have already been consumed.

**Expected behavior**
I expect that if materials are no longer available because they have been consumed by another craft, the indicators will update to reflect this.  In the above example to reproduce, the Darksteel Rivets would lose their blue background when the Darksteel Rings are marked completed, since they cannot be crafted without acquiring another Darksteel Nugget.

**Screenshots**
Example Reproduction

**Desktop (please complete the following information):**
 - OS: Windows 10 Home
 - Desktop app v7.5.6

AlaunusLux:
#1721 

This seems to follow the requested (and, to me, logical) behavior in the bug report. I have only tested with b-sen's example and a small example of my own, so it likely breaks something elsewhere. 

If nothing is broken and this is an acceptable fix, I'd still like to polish it up some more - I'm sure we could be more specific about which items to update, etc. Just wanted to see if I'm headed in the right direction, I guess!

image

### DIFF
--- a/apps/client/src/app/modules/list/model/list.ts
+++ b/apps/client/src/app/modules/list/model/list.ts
@@ -433,7 +433,6 @@ export class List extends DataWithPermissions {
     // If it's not a craft, break recursion
     if (getItemSource(item, DataType.CRAFTED_BY).length === 0 || item.requires === undefined) {
       // Simply return the amount of the item being equal to the amount needed.
-      return item.done >= amount;
       return item.done - item.used >= amount; //subtracting item.used here probably does something???
     }
     // If we already have the precraft done, don't go further into the requirements.
--- a/apps/client/src/app/modules/list/model/list.ts
+++ b/apps/client/src/app/modules/list/model/list.ts
@@ -332,6 +332,7 @@ export class List extends DataWithPermissions {
     if (item.done < 0) {
       item.done = 0;
     }
+    this.updateAllStatuses(item.id); //propagate changes
     const newDone = item.amount_needed - MathTools.absoluteCeil((item.amount - item.done) / item.yield);
     amount = newDone - previousDone;
     if (item.requires !== undefined && newDone !== previousDone) {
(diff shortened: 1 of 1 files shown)


Supamiu:
Looks good to me so far 🙂 

I was pretty sure an external look at this would have been better than mine, since you have a fresh context while I see this part as quite complex due to how badly it's designed in my head.

Thanks a lot, let me know if you need any details.

AlaunusLux:
There certainly is a lot of complexity! And I definitely don't understand it all, so I appreciate you taking a look :)

Supamiu:
Let me know when you think it's ready to be reviewed to merge :)

AlaunusLux:
Apologies for having fallen off on this, I'm sure you don't appreciate the hanging pull request.

I'd really like to reexamine it, but it's difficult to get back into it properly. If you'd like, we could merge what's here, or please feel free to improve on what I have if you feel that's best. 

I just can't guarantee when I'll be able to get back into it, and would hate to keep the app in a worse state for no real reason!

Another alternative would be removing the draft pull request so it's not hanging here in the open pull requests tab.

Let me know !

Supamiu:
If you want to take some time with me to talk about it or ask some questions about how the core if this works, we can setup that, in either english or french depending on the language you prefer and know ^^

AlaunusLux:
I mostly meant that I didn't have the free time/wasn't able to commit to it. But I've been watching some stuff that makes me want to take another crack at it. 

"You aren't done when it works, you're done when it's clean." - Uncle Bob

Supamiu:
I'm closing this for inactivity for now, let me know if you want to bring it back 👍

PR #2231, opened 2022-03-29
feat(data-extraction): update recipes format for zh/kr

zhyupe:


Supamiu (approved):

data-extraction/data-exporter/task/index.js
```diff
@@ -94,52 +94,61 @@ output('quests', () => db('Quest').simpleObject('Name'))
...
+      ingredients: ingredients.map(({ id, amount, ilvl }) => ({
+        id,
+        amount,
+        quality: canBeHq[id] ? (ilvl / totalIlvl * totalContrib) : null,
+      })),
+      progressDivider: +level.ProgressDivider,
+      qualityDivider: +level.QualityDivider,
+      progressModifier: +level.ProgressModifier,
+      qualityModifier: +level.QualityModifier,
+      expert: recipe.IsExpert === '1',
```
Supamiu, line 147: iirc IsExpert is a number not a string, which makes this condition always false, not a big deal but worth a fix later on.

PR #2505, opened 2023-02-17
chore(desktop): added some korean translation

kokose1234:


Supamiu:
Hi, thanks for your contribution.

Please prefer https://crowdin.com/project/ffxiv-teamcraft for translations as it'll make it follow the usual i18n flow which is better for me to integrate in the app.

PR #2518, opened 2023-02-23
fix(desktop): prevent open-link ipc event from opening local files

Scrxtchy:


Supamiu:
Hi, thanks for your PR.

I'm not sure what's your goal with this and it's hard to guess it from an empty description ^^"

Can you explain why you'd like to prevent open-link from opening local files and why you did it like this please?

Supamiu:
Thanks, it's much easier to understand now 👍 .

Having more security is definitely something cool and for that I have a whitelist of allowed ipc channels defined in preload.ts.

`open-link` being a TC-defined event, there's a very low chance of any external script using it to load something it shouldn't, but you're right, more security is never too much security, especially with externally loaded stuff.

If you want to dive deeper into this, I'll be happy to discuss it using text or voice on the Discord server (which I see you're already in). I know the entire codebase by heart and I'll be able to tell you if you can do something without breaking the app or not 😄 

I'm merging this, lgtm.

Supamiu (approved):

PR #2584, opened 2023-04-12
feat/voyage tracker UI enhancements

snb:
This is two small changes to show a little more info in the voyage tracker.

1) The abbreviated build in the build column is now colored to give an indication of the worst condition on the parts so it's easier to tell when a vehicle needs to be repaired without having to mouse over the build and look at the popup. It would be possible to instead color each part letter separately but that was more effort, would perhaps be enough extra color to be distracting, and it's generally sufficient to know the worst part condition because parts with higher than 0 don't need to be repaired anyway. Still, I'm not thrilled with how this looks and maybe it would be better to have a vertical red/green bar kind of icon like we see in game shown next to the build abbreviation.
2) Added a popover to the ranks column to show current / next level experience points, and a progress bar, for vehicles that aren't at max rank. When leveling a vehicle it can be convenient to know how close you are to reaching a new rank and I don't think this information was displayed anywhere in Teamcraft previously.

Please let me know if I have made silly errors or done things in a strange way. I'm not very familiar at all with TypeScript/Angular/any front-end web stuff so I kind of just tried to figure it out as I was going here.

### DIFF
--- a/apps/client/src/app/pages/voyage-tracker/voyage-tracker/vessel-row/vessel-build-column/vessel-build-column.component.html
+++ b/apps/client/src/app/pages/voyage-tracker/voyage-tracker/vessel-row/vessel-build-column/vessel-build-column.component.html
@@ -3,7 +3,7 @@
        [nzPopoverTitle]="name"
        [nzPopoverContent]="vesselBuildDetails"
   >
-    {{ vesselBuild?.abbreviation }}
+    <div [ngClass]="conditionClass()" class="condition">{{ vesselBuild?.abbreviation }} </div>
   </div>
 
   <ng-template #vesselBuildDetails>
--- a/apps/client/src/app/pages/voyage-tracker/voyage-tracker/vessel-row/vessel-build-column/vessel-build-column.component.html
+++ b/apps/client/src/app/pages/voyage-tracker/voyage-tracker/vessel-row/vessel-build-column/vessel-build-column.component.html
@@ -3,7 +3,7 @@
        [nzPopoverTitle]="name"
        [nzPopoverContent]="vesselBuildDetails"
   >
-    <div [ngClass]="conditionClass()" class="condition">{{ vesselBuild?.abbreviation }} </div>
+    <div [ngClass]="conditionClass$ | async" class="condition">{{ vesselBuild?.abbreviation }} </div>
   </div>
 
   <ng-template #vesselBuildDetails>
--- a/apps/client/src/app/pages/voyage-tracker/voyage-tracker/vessel-row/vessel-build-column/vessel-build-column.component.ts
+++ b/apps/client/src/app/pages/voyage-tracker/voyage-tracker/vessel-row/vessel-build-column/vessel-build-column.component.ts
@@ -7,7 +7,7 @@ import { VesselType } from '../../../../../modules/free-company-workshops/model/
 import { observeInput } from '../../../../../core/rxjs/observe-input';
 import { combineLatest } from 'rxjs';
 import { safeCombineLatest } from '../../../../../core/rxjs/safe-combine-latest';
-import { switchMap } from 'rxjs/operators';
+import { map, switchMap } from 'rxjs/operators';
 
 @Component({
   selector: 'app-vessel-build-column',
@@ -31,6 +31,16 @@ export class VesselBuildColumnComponent {
   @Input() parts: Record<AirshipPartType, VesselPart> | Record<SubmarinePartType, VesselPart>;
 
   parts$ = observeInput(this, 'parts');
+
+  conditionClass$ = this.parts$.pipe(
+    map(parts => {
+      const worstCondition = Math.min(...Object.keys(parts).map((slot) => (parts[slot].condition || 0) / 300));
+      return {'good': worstCondition > 50,
+              'caution': worstCondition <= 50 && worstCondition > 25,
+              'warn': worstCondition <= 25 && worstCondition > 0,
+              'broken': worstCondition <= 0}
+    })
+  )
 
   fullNameParts$ = combineLatest([
     this.parts$,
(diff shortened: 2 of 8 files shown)


apps/client/src/app/pages/voyage-tracker/voyage-tracker/vessel-row/vessel-rank-column/vessel-rank-column.component.ts
```diff
@@ -0,0 +1,32 @@
...
+@Component({
+  selector: 'app-vessel-rank-column',
+  templateUrl: './vessel-rank-column.component.html',
+  styleUrls: ['./vessel-rank-column.component.less'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class VesselRankColumnComponent {
+  @Input() rank: number;
+
+  vessel$ = observeInput(this, 'rank');
```
Supamiu, line 14: Don't observe input if you're not going to use the observable 😃 my guess is that you tried to replicate what you saw in other components (which is fine !) because as you said, you're not very familiar with Angular.

apps/client/src/app/pages/voyage-tracker/voyage-tracker/vessel-row/vessel-build-column/vessel-build-column.component.html
```diff
@@ -3,7 +3,7 @@
        [nzPopoverTitle]="name"
        [nzPopoverContent]="vesselBuildDetails"
   >
-    {{ vesselBuild?.abbreviation }}
+    <div [ngClass]="conditionClass()" class="condition">{{ vesselBuild?.abbreviation }} </div>
```
Supamiu, line 6: Calling a function in Angular template binding is a bad practice because the change detector will have to call it everytime it "travels" here to see if previous value is == new value, in order to know if yes or not, the binding should be updated.

I'll put a code suggestion in the TS part of it to explain how to implement this differently.

apps/client/src/app/pages/voyage-tracker/voyage-tracker/vessel-row/vessel-build-column/vessel-build-column.component.ts
```diff
@@ -58,4 +58,12 @@ export class VesselBuildColumnComponent {
...
     return Object.keys(this.parts).map((slot) => `${((this.parts[slot].condition || 0) / 300).toFixed(2)}%`).join(' - ');
   }
+
+  conditionClass(): object {
+    const worstCondition = Math.min(...Object.keys(this.parts).map((slot) => (this.parts[slot].condition || 0) / 300));
+    return {'good': worstCondition > 50,
+            'caution': worstCondition <= 50 && worstCondition > 25,
+            'warn': worstCondition <= 25 && worstCondition > 0,
+            'broken': worstCondition <= 0}
+  }
```
Supamiu, line 68: Instead of this, you'll want to use an `Observable`, because observables are only triggered when the value changes, making your code efficient for this, because you don't need to update condition class if parts didn't change at all.

Something like:

```ts
conditionClass$ = this.parts$.pipe(
                                              map(parts => {
                                                  const worstCondition = Math.min(...Object.keys(this.parts).map((slot) => (this.parts[slot].condition || 0) / 300));
                                                  return {'good': worstCondition > 50,
                                                              'caution': worstCondition <= 50 && worstCondition > 25,
                                                              'warn': worstCondition <= 25 && worstCondition > 0,
                                                              'broken': worstCondition <= 0}
                                                  })
                               );
```

Right after the `parts$` observable and then use it in template by doing `conditionClass$ | async` instead of `conditionClass()`.

apps/client/src/app/pages/voyage-tracker/voyage-tracker/vessel-row/vessel-rank-column/vessel-rank-column.component.ts
```diff
@@ -0,0 +1,32 @@
...
+  totalExperienceForNextRank$ = observeInput(this, 'totalExperienceForNextRank');
+
+  @Input() maxRank: number;
+
+  constructor(private freeCompanyWorkshopFacade: FreeCompanyWorkshopFacade) {
+  }
+
+  rankUpPercent(): number {
+    return Math.round(100 * (this.currentExperience / this.totalExperienceForNextRank));
+  }
```
Supamiu, line 31: You should make use of these nice `Observable`s above instead of making a method to then call it in template (as explained above, bad practice etc).

I'm going to write the `Observable` here with comments so you can understand it nicely:

```ts
// combineLatest because we want to only compute again if one of the observables emitted a new value, combined 
// with the latest value from the other one
rankupPercent$ = combineLatest([this.currentExperience$, this.totalExperienceForNextRank$]).pipe(
  // map will allow us to project these two values into tyhe computed value that we want.
   map(([currentExperience, totalExperienceForNextRank]) => {
    // Your original formula, but this time it'll only be executed if one of the inputs changes, not every frame that the change detector will compute.
    return Math.round(100 * (currentExperience / totalExperienceForNextRank));
   })
)
```

Then you can use it by doing `rankupPercent$ | async` instead of `rankupPercent()` in the template.

Remember to delete the `Observable` fields that you're not using to avoid creating `Observable`s for nothing 🙂.

Supamiu (changes requested):
For someone who was fiddling with something they do'nt know, your code is good 👍 just missing some good practices but they're specific to Angular and RxJS here, so that's ot be expected 😄

snb:
Thanks so much for the feedback! I'll try to make those changes soon

Supamiu (approved):

Supamiu:
You did an amazing job fixing the code, thanks for your contribution ! It'll be included in the next Teamcraft update.

PR #2601, opened 2023-05-07
fix(custom-alarm-without-weather): custom alarm without weather are now properly displayed

Jengines:
Adding a default value for weather on custom alarm form Allow alarm with missing spawns and remaining time to be displayed

### DIFF
--- a/apps/client/src/app/core/alarms/+state/alarms.facade.ts
+++ b/apps/client/src/app/core/alarms/+state/alarms.facade.ts
@@ -284,7 +284,8 @@ export class AlarmsFacade {
   }
 
   public createDisplayArray<T extends AlarmDetails | PersistedAlarm = PersistedAlarm>(alarms: T[], date: Date): AlarmDisplay<T>[] {
-    return this.sortAlarmDisplays(alarms.map(alarm => {
+    return this.sortAlarmDisplays(alarms.filter(alarm => alarm.spawns !== undefined || alarm.weathers !== undefined)
+      .map(alarm => {
         return this.createDisplay(alarm, date);
       }));
   }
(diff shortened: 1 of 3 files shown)


apps/client/src/app/core/alarms/+state/alarms.facade.ts
```diff
@@ -284,8 +284,7 @@ export class AlarmsFacade {
   }
 
   public createDisplayArray<T extends AlarmDetails | PersistedAlarm = PersistedAlarm>(alarms: T[], date: Date): AlarmDisplay<T>[] {
-    return this.sortAlarmDisplays(alarms.filter(alarm => alarm.spawns !== undefined || alarm.weathers !== undefined)
```
Supamiu, line 287: Removing this condition means that we can have alarms that have no spawn nor weather, which means they are always up, which raises the question: what's the point of having an alarm that's 100% up?

Supamiu (changes requested):

apps/client/src/app/core/alarms/+state/alarms.facade.ts
Jengines: Was aking myself the exact same question. But it's actually possible to create a custom alarm without indicated a spawn (and the alarm is created), so I decided it would be better to now saw them on alarms page.

A lot of users may have created alarms without spawn and never saw them (but they still exists in firebase, was my case).

If we decide to make the spawn needed, we should make the field required on custom alarm form, and having a way to remove (or update) existing alarms in firebase who don't have a spawn.

What's your feeling about that : 
 - Should we still allow alarms without spawn (but for what usage... I don't know)
 - Or decide to make the spawn required on custom alarm form and how you think it's possible to update inconstitent data on firebase ?

Supamiu: I don't think we should take existing data into consideration at all tbh, I'd say the form should have a global validator that checks if either weather or spawn is filled, this way we can make them optional but you have to fill at least one of them.

See https://angular.io/guide/form-validation#adding-cross-validation-to-reactive-forms for guidance on how to do it.

Supamiu (approved):

PR #2610, opened 2023-05-20
chore(i18n): Dutch-translations

JenTheBluePanda:
I started doing some translations for Dutch.

Sorry, I didn't realize the project was on Crowdin until I tried to open this PR, as it wasn't linked in the README or CONTRIBUTING guide.

However, would it be fine to continue translating via Git? I noticed your setup with Crowdin does not attribute translators in `git log` or GitHub, unfortunately.


SethFalco:
I was looking at this with Jen, so just wanted to chime in with my thoughts!

I was reading the Crowdin documentation, but it looks like they really don't support preserving authorship, which is unfortunate. :thinking: 

It'd be great if the flow was more like Weblate. Weblate supports git integration as well, but preserves authorship/attribution of contributors, which is required by most open-source licences, including MIT, anyway.

> The author of every commit (if the underlying VCS supports it) is still recorded correctly as the user that made the translation.
> 
> — https://docs.weblate.org/en/latest/faq.html#why-are-all-commits-committed-by-weblate-noreply-weblate-org

Basically, Weblate does something to the effect of:
`git commit -m "Translations" --author "Name <name@example.org>"`

Which makes the commit author the translator, but the committer the bot.

image
> Example of this in the [Jellyfin](https://github.com/jellyfin/jellyfin/commits/master) repository.

Supamiu:
Sadly I'm not able to change the way crowdin makes translation PRs, it's easier to manage translations using it because it doesn't require contributors to have a github account at all. Which is why moving to a new system is not really ideal imho.

If you want to be credited for your work on github, keep using PRs instead but you'll probably have conflicts with new translations being added on new features, shouldn't be too hard to resolve tho.

Supamiu (approved):

PR #2641, opened 2023-07-13
fix(accessibility): add lang attribute to html

SethFalco:
This PR sets `html[lang]` with the current language, and updates it if the language is changed while in the SPA.

## Motivation

On the web it helps browsers, translation tools, and screen readers if the language of the content is defined in `html[lang]`.

For example:

* Translation tools know what to translate the page from without user input.
* Screen readers know what language/accent to use when reading content.

## References

> This Success Criterion helps:
>
> * people who use screen readers or other technologies that convert text into synthetic speech;
> * people who find it difficult to read written material with fluency and accuracy, such as recognizing characters and alphabets or decoding words;
> * people with certain cognitive, language and learning disabilities who use text-to-speech software
> * people who rely on captions for synchronized media.
>
> — [Web Content Accessibility Guidelines: Language of Page](https://www.w3.org/WAI/WCAG21/Understanding/language-of-page)

> WCAG Success Criterion 3.1.1 requires that a page language is specified in a way which may be 'programmatically determined' (i.e. via the lang attribute).
> …
> The purpose of these requirements is primarily to allow assistive technologies such as screen readers to invoke the correct pronunciation.
> 
> — [MDN Web Docs: lang attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang)

### DIFF
--- a/apps/client/src/app/app.component.ts
+++ b/apps/client/src/app/app.component.ts
@@ -47,7 +47,7 @@ import { CustomItemsFacade } from './modules/custom-items/+state/custom-items.fa
 import { DirtyFacade } from './core/dirty/+state/dirty.facade';
 import { SeoService } from './core/seo/seo.service';
 import { Theme } from './modules/settings/theme';
-import { isPlatformBrowser, isPlatformServer } from '@angular/common';
+import { DOCUMENT, isPlatformBrowser, isPlatformServer } from '@angular/common';
 import { REQUEST } from '@nguniversal/express-engine/tokens';
 import * as semver from 'semver';
 import { UniversalisService } from './core/api/universalis.service';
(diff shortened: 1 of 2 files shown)


SethFalco:
> FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
> 
> — https://app.circleci.com/pipelines/github/ffxiv-teamcraft/ffxiv-teamcraft/8246/workflows/27090742-7c4c-4b80-902d-c307bf5f2939/jobs/42522

Hmm, thinking this is unrelated to my changes.

Supamiu:
Hi, thanks for your contributions !

Heap error is indeed unrelated, I'll fix this eventually 😄.

Using the renderer was a good idea but I'd like to avoid using selectors like that that will do a full DOM lookup to grab an element that's pretty much static.

I think the better approach would be to use `DOCUMENT` ref, as shown in this SO reply: https://stackoverflow.com/a/76632181/4102561

SethFalco:
> I think the better approach would be to use DOCUMENT ref, as shown in this SO reply: https://stackoverflow.com/a/76632181/4102561

I didn't know about this! Good suggestion.
I've tested this solution, and amended the PR to take that approach instead.

Supamiu:
Amazing !

Issue #2686, opened 2023-08-21
bug: use-durability-restoration-later tip does not adjust index for filtered steps


phx-evan-hilscher:
The following code in use-durability-restoration-later.ts retrieves the steps during which Manipulation is active:

`const repairSteps = simulationResult.steps.filter(s => !s.action.skipsBuffTicks()).slice(index + 1, index + 1 + new Manipulation().getDuration(simulationResult.simulation));`

In https://github.com/ffxiv-teamcraft/ffxiv-teamcraft/commit/a8a36b4af0ca05d7ea7d41b6005669f1c6321c3b, a filter was added to ignore steps that do not advance the buff tick (such as Final Appraisal):

`simulationResult.steps.filter(s => !s.action.skipsBuffTicks())`

However, if steps are found and filtered out, this will change the desired slice bounds. ~~The number of steps filtered must be counted and subtracted from the slice bounds.~~

~~`const filtered =  simulationResult.steps.filter(s => !s.action.skipsBuffTicks())`
`const filterCount = filtered.length`
`const repairSteps = filtered.slice(index - filterCount + 1, index - filterCount + 1 + new 
Manipulation().getDuration(simulationResult.simulation));`~~

The solution proposed here is not quite sufficient.

Erroneous tip:
![image](https://github.com/ffxiv-teamcraft/ffxiv-teamcraft/assets/102170871/f6b213ad-5b97-45b2-be09-327bf60db670)

Same rotation, but with Final Appraisal removed:
![image](https://github.com/ffxiv-teamcraft/ffxiv-teamcraft/assets/102170871/dec48297-94f5-44ed-8e98-b9aea694d0b2)

Supamiu:
Hi, that's a nice report despite it not following the template, I won't instaclose this one 😄 

As I see you did most of the investigation, it'd make sense to have you suggest the fix as a PR 🙂 I'll be happy to help you through that but basically, everything is in CONTRIBUTING.md to get you started on how to run local Teamcraft, how to contribute, it should be pretty easy for you.

PR #2687, opened 2023-08-21
fix(simulator): Fix logic for manipulation check in use-durability-restoration-later tip

phx-evan-hilscher:
See issue https://github.com/ffxiv-teamcraft/ffxiv-teamcraft/issues/2686, bug: use-durability-restoration-later tip does not adjust index for filtered steps


Supamiu:
Hey, nice PR ! Thanks !

Did you test it against local simulator? Did it match the expected behavior?

phx-evan-hilscher:
Yes, I did. It works as expected now for the handful of test cases I came up with.

Supamiu (approved):

PR #2698, opened 2023-09-17
fix(log-tracker): Ensure correct page is displayed when 'Hide Completed' is enabled.

### LINKED ISSUE #2061: bug:  Log Tracker unable to open First Section in a list when Hide Completed is used
**Describe the bug**
When you use the Hide Complete feature on the Log Tracker (Helpers -> Log Tracker) the first Section (by section I mean the List of Groupings such as Housing (1) or Housing (2) found on the Left side of the screen) is not able to be open when you go to a New List (i.e. CRP, MIN, BTN, etc.) 

**To Reproduce**
1) Go to the Log Tracker 
2) Turn Hide Completed Items on 
3) Go to a New List i.e. CRP, MIN, BTN, ETC
4) Attempt to click on the First Section in the List 
5) Unable to see the items in the First Section
WORKAROUND: Click on a different Section i.e. Housing (2) instead of Housing (1) and then click to the original Section and it works. This Workdaround does not work if ONLY one section is left on a list.

**Expected behavior**
You should be able to see into the first section of a list without needing to click on another section first.

**Screenshots**
image
image
image
image

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser Desktop App
 - Version 9.0.3

Judgy53:
Fixes #2061 partially. I noticed 2 scenarios where it still happens : 

- Switching back and forth between 2 DoH jobs. The current page index of a job gets stored somehow, and it mess up the page display when switching back and forth with another job. I suspect it is an issue with the html template, as it doesn't happen with DoL sub tabs, but I can't pinpoint the exact spot.
- Somewhat randomly when switching between DoH and DoL. I think it's a lazy loading/race condition issue, but didn't investigate very far.

### DIFF
--- a/apps/client/src/app/pages/log-tracker/log-tracker/log-tracker.component.html
+++ b/apps/client/src/app/pages/log-tracker/log-tracker/log-tracker.component.html
@@ -10,26 +10,26 @@
             <div fxLayout="row" fxLayoutGap="10px">
               <div fxLayout="row" fxLayoutAlign="flex-end center" fxLayoutGap="5px">
                 <nz-switch [(ngModel)]="hideCompleted"
-                           (ngModelChange)="updateSelectedPage($event, showNotRequired, trackerType, dohTabs, [], true); selectedRecipes = {}"></nz-switch>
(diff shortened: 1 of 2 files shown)


Supamiu (approved):
Thanks for your contribution !

PR #2713, opened 2023-10-04
feat(add-rare-animals): add rare animals from 6.5 update

Jengines:
Adding rare animals static data


Jengines:
The information was extracted from Overseas Casuals Discord.

Two of them was verified IG. From the other I think the information are accurates (according to Discord fiability, and other site having the same info)

libs/data/src/lib/json/mobs.json
```diff
@@ -75756,28 +75756,28 @@
     "fr": ""
   },
   "12627": {
-    "en": "morbol",
```
Supamiu, line 75759: Did you make these changes by hand? because they're extracted from the game so these changes will be overriden next time the extractors are going to be used, which will be sooner than later for sure.

Jengines: Yes, I do it manually, I haven't check if it was extracted.

It's strange , the first letter is not capitalized, seeing this mob in game and it was. Maybe we can try to re-run the extractor ?

However, we can keep this change and it would be overwritted?

Supamiu: Yeah it'll just be overwritten, this is due to some small capitalization issue, probably not extracting from the right column but TC should handle this fine iirc.

PR #2717, opened 2023-10-11
fix(core): fix Japanese aetheryte names

katota1114:
For example, both `Gridania Aetheryte Plaza` and `Limsa Lominsa Aetheryte Plaza` translate into `エーテライト・プラザ` in Japanese, so Japanese users doesn't understand which one is `Gridania`.

So, I add city name before `エーテライト・プラザ` like English.


Supamiu:
Hi, thanks for your contribution

Unfortunately what you edited are files that are generated based on game files, meaning that the next time the extractor will be run, your changes will be overriden. Your issue is due to how the game stores names for these aetherytes and I'd say maybe it'd be better to fix that in an `aetheryteName` pipe but not even because it'd mean implementing a per-language logic, which is an issue.

I can't accept this PR because it'll be overriden with every extractor run, sorry.

Issue #2479, opened 2023-02-04
feat: ignore tribal vendors in lists


lucasbfr:
**Is your feature request related to a problem? Please describe.**
When creating a list for items to craft, teamcraft suggests items that are unavailable if you didn't unlock tribes, which makes it harder to spot the items that I need to gather in realty. This is especially an issue when using step by step mode (these gatherings don't appear in the list because they can be ignored by someone able to buy the item)

**Describe the solution you'd like**
A way either to tell TC to ignore a specific vendor, or to ignore all tribal vendors when in a list.

**Describe alternatives you've considered**
I usually take a note of the items that are marked as "can be ignored" in notepad so that I don't forget to gather them when I'm in the area

lucasbfr:
Awesome! I had no idea layout existed. I confirm that this works:
![2023-02-04 19_29_08-sharlayan 80 — Mozilla Firefox](https://user-images.githubusercontent.com/2010753/216783655-97375a68-4f72-403f-adbb-e990fce7e5bc.png)

However layouts are really for power users so a button in the vendor panel to toggle this behaviour would be great

anilanar:
I don't think this works for "Sylphic Vendor" that seels , which means I need to farm sylphic reputation to unlock this NPC, which is a bummer.

Supamiu:
This might be due to this specific vendor not being in the handmade list of tribe vendors, I'd need to have its ID in order to check.

PR #2767, opened 2024-02-10
feat(layouts): Add CAN_BE_BOUGHT_WITHOUT_TRIBES layout filter

Daniel324225:
I added this filter because the solution in #2479 doesn't work properly, since some items can be bought from tribe vendors as well as normal vendors.

CAN_BE_BOUGHT AND NOT FROM_BEAST_TRIBE:
image
CAN_BE_BOUGHT_WITHOUT_TRIBES:
image


Supamiu (approved):

Supamiu:
Thanks a lot, this is a very nice contribution !

PR #2828, opened 2024-06-16
[WIP] chore(Debugging): VSCode Configuration for Debugging Electron Renderer

forteddyt:
Introduces a VSCode configuration to support debugging the electron renderer process in VSCode. Based heavily off of https://github.com/abartho/electron-typescript-vscode

What this effectively means is that starting the "Electron: All" configuration enables breakpoints set in vscode to be hit


forteddyt:
The current, WIP PR, does not seem to associate source maps properly. The renderer process successfully attaches to the main process (console logs show in the Electron: Renderer debug window), but breakpoints do not seem to get hit. To me, this seems like a source map issue but I'm not too certain

forteddyt:
one difference I am noticing between the repo I'm basing my launch configuration is that there is not a source map file for the electron main.js file:
image

whereas the reference repo does have a mapping:
image

Supamiu:
This is due to sourceMap being set to false for the build config, see https://github.com/ffxiv-teamcraft/ffxiv-teamcraft/blob/0d40c832c5c9c661a0c4be48907fbe079addd114/apps/electron/tsconfig.app.json#L8

Feel free to change that in your PR.

forteddyt:
Ya... I eventually found that, as well. but a couple things:
1. adding sourceMap: true to electron/tsconfig.app.json results in the generation of only dist/apps/electron/src/preload.js.map
2. adding sourceMap: true to electron/project.json results in the generation of only dist/apps/electron/main.js.map
image

However, even with both of these source maps, the attached renderer still does not hit breakpoints. I'll still push up these changes, because it feels like this is on the right track

I will say I was expecting modifying tsconfig.app.json would have generated source maps for all of the electron files, so that's kind of weird...

The last note I'll make is that when running the Electron: Renderer config, I get the following warning (yellow text): `Could not load source map from preload.js.map`. Part of me is wondering if that is an entrypoint for the source mapping somehow, and since that is broken, the rest breaks as well. I tried looking around online for such an error message, and the general sentiment seems like that message is just a warning and can be ignored.

Supamiu:
preload is really only used for few things and isn't worth adding debugger imo

Supamiu:
Hey, any status update on this?

forteddyt:
No, unfortunately

I gave it a few more hours of googling and trial and error but was not able to achieve what I was aiming to. I haven't had much free time since then, and my limited knowledge of Electron has essentially killed my ambition to implement this PR 😅 

Feel free to close the PR, if that's why you're asking. If I eventually get back around to it, I can just re-open

PR #2873, opened 2024-07-11
feat(simulator): option to turn off simulation constraints for quick building of macros

Kosyne:
Teamcraft's simulation tools are fairly sophisticated, but sometimes you just want to make a quick and dirty macro by just clicking some skill icons then exporting to FFXIV. (particularly for users without accounts) There are a few tools out there that do this, but they're generally outdated. This enables that kind of click-and-export simplicity inside TC by means of a optional toggle at the top of the crafting simulation page.

Upon clicking it, the simulation parts of the page are hidden, validations are ignored, and the user is allowed to simply select the skills they want and export. Clicking it again reverts back to the original behavior.

First time contributing to TC, so let me know if there's any loose ends and I'll be more than happy to tie them up.

### DIFF
--- a/apps/client/src/app/pages/simulator/components/simulator/simulator.component.html
+++ b/apps/client/src/app/pages/simulator/components/simulator/simulator.component.html
@@ -270,6 +270,14 @@
             </div>
             <div>{{ 'SIMULATOR.Safe_mode' | translate }}</div>
           </div>
+          <div [nzTooltipTitle]="'SIMULATOR.Macro_only_mode_tooltip' | translate" fxLayout="row" fxLayoutAlign="flex-start center"
+               fxLayoutGap="5px"
+               nz-tooltip>
+            <div>
+              <nz-switch [ngModel]="(macroOnlyMode$ | async)" (ngModelChange)="macroOnlyMode$.next($event)"></nz-switch>
+            </div>
+            <div>{{ 'SIMULATOR.Macro_only_mode' | translate }}</div>
+          </div>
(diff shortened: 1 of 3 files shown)


Supamiu:
I feel like this shouldn't be the use case for simulator, I'd rather have a separate page for quick macro building with added things such as micon, sounds, etc, also because I expect people to ask for them at some point.

Supamiu:
Do you want me to create a page for this so you can implement a macro builder instead?

Kosyne:
Hm, I agree that's definitely good way to go about it, and as a user I would love to have such a page. Unfortunately I can't guarantee I have the time to implement that at present, hence the quick piggybacking off of an existing page. That said, I can certainly look into what all may be involved to implement that (still getting a feel for the project, best practices, etc).

From your wording, is making a page something that'd need to be done on your end, or something I can can stand up locally to test?

Supamiu:
I can do it for you if you're not sure how to do it (nx generate) or where to connect it (in router) and you want to just have a blank page to start from.

If oyu want to give it a go and add a new page, I'm fine with it too 🙂

Kosyne:
What would you consider the a MVP here? An entry in the left navbar, a job selector containing just crafting classes (initially), and icons relevant to those jobs, probably reusing the macro builder code from the sim page where it can, or were you wanting to go further / different?

Supamiu:
Yeah, I'm not even sure job selection would be useful tbh because they all have the same action names anyways and that's the important part for macro. Could even have macro build as the user inputs skills and split into chunks too.

PR #3030, opened 2025-05-10
refactor(maps): update cosmic exploration maps to use XIVAPI v2

luan-dev:


Supamiu:
These files are generated, editing them by hand is never a solution 🙂 .

Instead, you'll want to change how it's generated here:

https://github.com/ffxiv-teamcraft/ffxiv-teamcraft/blob/a81ea73d705205d64965b9eddf500cecca5088f2/apps/data-extraction/src/extractors/maps.extractor.ts#L25

Also, I just uploaded the missing maps on legacy xivapi, should be good to go (I completly forgot about them, yup)

luan-dev:
oh shoot. thanks for the clarification. i'll keep that in mind next time

PR #3129, opened 2026-01-31
update for cn patch-7.4

mourenhe:
update for cn patch-7.4


Supamiu:
7.4 contains more than just these items and I think the CN community has a datamining worker supposed to open a PR?

mourenhe:
> 7.4 contains more than just these items and I think the CN community has a datamining worker supposed to open a PR?
Sorry for the late reply. I’m just a regular player with no programming background, and all these updates were done with the help of AI. What I updated are the entries with English names in the file libs/data/src/lib/json/items.json, and the last ID in this file is 50852—this is the extent of what I can do. I’m sorry, but I have no idea where to start with the additional data updates for 7.4. Prior to 7.4, there was a dataminer working on this, and you can see that many files under the path libs/data/src/lib/json/zh haven’t been updated since patch-7.35, with no new content added for 7.4.
image

mourenhe:
> 7.4 contains more than just these items and I think the CN community has a datamining worker supposed to open a PR?

Before this PR, I submitted the one titled feat(items): add i770 item zh CN translation #3119 as an initial attempt. After the Patch 7.4 update, the previous dataminer ceased further updates. When using the list function in Teamcraft, some item names only displayed in English—this made it impossible to copy them for searching and also looked highly unintuitive, which is why I made that earlier attempt.

PR #3180, opened 2026-04-19
feat: use TW version's data in search and db pages

augenfrosch:
Since packet capture might soon be added for the TW version, I took a look at what is missing for the search and database pages to work for it.

The changes are just copy-paste from what is done for `zh` and `ko`. (There are actually more places where `tw` isn't being accounted for in the same way, but I didn't have an easy way of testing if changes to those would work, and didn't want to bloat the scope of this PR.)

I'm not including the generated files in the PR since the db page creation uses `SaintCoinach.Cmd`, which only runs on Windows. I did test it with some modifications to use CSV files generated by [XIVData Oxidizer](https://github.com/skyborn-industries/xiv-data-oxidizer) though.

screenshot showing the search working with TW selected

screenshot showing an item db page with TW selected

`./assets/icons/taiwan.png` currently 404s since it doesn't exist yet.
(The craft type and patch text not loading in is seemingly an issue for all non-global versions / also happens on the main site for `KO` and `ZH`)

### DIFF
--- a/apps/client/src/app/core/tools/normalize-i18n.ts
+++ b/apps/client/src/app/core/tools/normalize-i18n.ts
@@ -1,7 +1,7 @@
 import { I18nName, XivapiI18nName } from '@ffxiv-teamcraft/types';
 
 export function normalizeI18nName(row: I18nName | { name: I18nName } | XivapiI18nName): I18nName {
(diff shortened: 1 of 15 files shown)


Supamiu (approved):

Supamiu:
Amazing work, thanks a lot!

PR #3197, opened 2026-04-30
feat(linux): Add initial support for Linux, building as an AppImage

birkholz:
[Draft because work-in-progress]

This adds initial support for a native desktop version of Teamcraft's electron app.

This depends on MRs for [pcap-ffxiv](https://github.com/ffxiv-teamcraft/pcap-ffxiv/pull/46) and [dll-inject](https://github.com/Supamiu/dll-inject/pull/8) which need to be merged first.

Because the game runs in Wine on Linux, and its not possible to connect to a process in Wine from outside of Wine, [deucalion-bridge](https://github.com/birkholz/deucalion-bridge) was created. It is a small Rust server that Teamcraft runs in Wine alongside the game. It does the injection of deucalion's DLL, and then creates a bidirectional TCP bridge between deucalion and Teamcraft that otherwise acts the same as packet capture on Windows. Teamcraft manages all of this with no action required from the user.

There are a few required configurations in order to get packet capture working. We need to know:
* Where the game stores user config files
* Where the user's wine executable is
* Where the user's wineprefix/protonprefix is

The user is required to provide these via the settings menu before packet capture can be enabled.

Just as in Windows, the game must be running before enabling packet capture.

### Known Issues
* Overlays
  * transparency does not function, may be added in a future MR
  * "always on top" may work, but depends on the user's window manager
* Auto-update does not work

### Planned package formats
* AppImage
* DEB
* RPM

Not supported:
* Flatpak
* Snap

Because these package formats sandbox the app, making packet capture impossible.


package.json
```diff
@@ -3,7 +3,7 @@
   "description": "Collaborative crafting made easy",
   "homepage": "https://ffxivteamcraft.com",
   "author": "FFXIV Teamcraft",
-  "version": "11.4.22",
+  "version": "11.4.22-linux",
```
Supamiu: Why adding linux suffix to this?

package.json
```diff
@@ -31,18 +31,30 @@
         "teamcraft"
       ]
     },
+    "publish": {
+      "provider": "github",
+      "owner": "birkholz",
+      "repo": "ffxiv-teamcraft",
+      "releaseType": "release"
+    },
```
Supamiu, line 39: Should probably make this match the official one if you want it merged ^^

package.json
```diff
@@ -109,14 +124,16 @@
   },
   "private": true,
   "dependencies": {
-    "@ffxiv-teamcraft/pcap-ffxiv": "1.2.30",
+    "@ffxiv-teamcraft/pcap-ffxiv": "birkholz/pcap-ffxiv#main",
```
Supamiu: I assume we'll need a new release pof pcap-ffxiv but better fix this too now

package.json
birkholz: Just a temporary measure while I build separately from the official version, so the version doesn't conflict and I can increment (`-linux2`) etc to test autoupdating while keeping it clear what official version its based on.

package.json
birkholz: Yup, this can be changed back to `1.2.31` or whatever once https://github.com/ffxiv-teamcraft/pcap-ffxiv/pull/46 is merged

It just needs a bit more work before its ready.

package.json
birkholz: Yup, was just temporary so I could publish builds to my repo while working on it.

.github/workflows/build-linux.yml
```diff
@@ -0,0 +1,33 @@
+name: Build Linux AppImage
```
ayyaruq: Good for testing but I think once this is ready for merge it should run on CircleCI like the rest of our release process, just to get it all in the proper pipeline sequence.

apps/client/src/app/app.component.ts
```diff
@@ -317,6 +317,12 @@ export class AppComponent implements OnInit {
       // Translation
       this.use(this.getLang());
       if (this.platformService.isDesktop()) {
+        this.ipc.on('checking-for-update', () => {
+          this.checkingForUpdate$.next(UpdaterStatus.DOWNLOADING);
+        });
```
birkholz, line 322: I noticed a bug where it was endlessly spinning saying it was downloading an update, even when there was no update available. The `checking-for-update` event was not in preload.ts.

birkholz:
Closing in favor of https://github.com/ffxiv-teamcraft/ffxiv-teamcraft/pull/3206 and a more focused Linux MR (currently https://github.com/birkholz/ffxiv-teamcraft/pull/1)

PR #3201, opened 2026-05-07
feat(electron): add macOS desktop build support

spidermonky0007:
## Summary

Adds initial macOS desktop packaging support for the Electron app.

## Changes

- Adds macOS Electron Builder scripts for `.app` and `.dmg` builds.
- Adds a macOS app icon for packaged builds.
- Fixes Electron local HTML loading by using `file://` URLs.
- Uses platform-safe path handling for the FFXIV documents folder while keeping existing Windows/KR/CN behavior intact.
- Adds XIV on Mac character data path detection for the DAT watcher.
- Adds a macOS packet capture unsupported state because the current packet capture tooling depends on Windows-only components.
- Adds English strings for the macOS packet capture unsupported message.

## Known limitations

- Packet capture is not supported on macOS yet.
- The current packet capture implementation depends on Windows-only Deucalion/dll-inject tooling.
- This PR intentionally makes packet capture fail gracefully on macOS instead of showing the misleading antivirus/missing injector message.

## Tested

- Built macOS arm64 `.app` successfully.
- Built macOS arm64 `.dmg` successfully.
- Confirmed packaged app launches.
- Confirmed app icon displays correctly.
- Confirmed login and session persistence.
- Confirmed search, lists, list creation, and account list data work.
- Confirmed DAT watcher resolves to the XIV on Mac config path containing `FFXIV_CHR...` folders.
- Confirmed packet capture shows the macOS unsupported message.
- Tested on macOS Tahoe arm64 only. Earlier macOS versions have not been tested yet.


spidermonky0007:
Converted to draft due to bug for char fodlers

apps/electron/src/main.ts
```diff
@@ -76,6 +76,10 @@ const ipcListenersManager = new IpcListenersManager(pcapManager, overlayManager,
 const squirrelEventHandler = new SquirrelEventHandler(pcapManager, store);
 const isUpdating = squirrelEventHandler.handleSquirrelEvent();
 
+if (process.platform === 'darwin') {
+  app.disableHardwareAcceleration();
```
birkholz, line 80: I assume this is because it performs poorly on some hardware? Can the user still enable it in the settings if they have a discrete GPU powerful enough for it?

spidermonky0007: Yes it's just off by default if on Mac, I have a pretty high end Mac and it was causing issues. They can enable by in settings still

spidermonky0007:
I’m leaving this draft here for reference, but this work has moved to the stacked PR based on birkholz:feat/cross-platform-shared.

The newer stacked PR builds on the shared cross-platform desktop/bridge work instead of duplicating it here.

The remaining macOS-specific pieces are now focused on:
- XIV-on-Mac DAT watcher path handling
- macOS fs.watch rename events
- allowing the bridge to start before the game on macOS
- staging Deucalion files into the XIV-on-Mac Wine prefix
- required XIV-on-Mac Wine env vars
- downloading bcryptprimitives.dll without committing the binary

Once the stacked PR direction is confirmed, I can close this older draft to avoid review confusion.

PR #3206, opened 2026-05-09
feat(pcap): add shared cross-platform desktop plumbing and bridge lifecycle

birkholz:
Adds the shared Electron changes needed by both the Linux (https://github.com/birkholz/ffxiv-teamcraft/pull/1 ) and the MacOS (https://github.com/birkholz/ffxiv-teamcraft/pull/2 ) MRs, including the bridge lifecycle. MRs will be re-targeted to staging after this is merged.

This does not add the new builds, only the requisite shared code to enable Linux/MacOS.

I wasn't sure what the preference would be around bundling the bridge executable, so I took the same approach as pcap-ffxiv does with the deucalion DLL.

### What does merging this do for Windows?

* `deucalion-bridge.exe` isn't bundled
* The desktop settings are unchanged
* None of the bridge lifecycle logic & events are loaded


apps/electron/src/dat/dat-files-watcher.ts
```diff
@@ -61,14 +61,21 @@ export class DatFilesWatcher {
       this.stop();
     });
 
+    ipcMain.on('dat:all-odr', (event) => {
```
birkholz, line 64: Pulled out of `start()`, fixed a bug on Linux

apps/electron/src/dat/dat-files-watcher.ts
```diff
@@ -61,14 +61,21 @@ export class DatFilesWatcher {
...
+      const watchDir = this.getWatchDir();
+      if (watchDir) {
+        event.sender.send('dat:all-odr:value', this.getAllItemODRs(watchDir));
+      }
+    });
+
     ipcMain.on('dat:path:get', (event) => {
-      event.sender.send('dat:path:value', this.getWatchDir());
+      event.sender.send('dat:path:value', this.getWatchDir() ?? '');
```
birkholz, line 72: It's possible for the default directory to be missing in Linux/Mac, so added some handling if the directory is missing.

package.json
```diff
@@ -24,6 +24,17 @@
         ]
       }
     ],
+    "linux": {
```
birkholz, line 27: We're only bundling the bridge executable in linux and mac builds, because it might trigger antivirus software in Windows.

PR #3229, opened 2026-06-03
Update Grand Company Supply data: Move 36240 from ALC 85 => ALC 86

mav8557:
Hi! Thanks for this project, it's great and I use it very often. 

I was working on my turn-ins and using the data file from this repository when I noticed that my current turn-in for my Level 86 ALC, Sungold Firesand was in the level 85 turn-in list. For example, here is what it looks like in game:

image

From my understanding this might not matter on the actual Teamcraft site, since it would still show the item in the drop-down list that I'm assuming is populated from this file. This PR moves the item to the level 85 block to the level 86 one.

That said, if this is not an error and I'm just reading this file wrong, feel free to reject this. Just figured I'd ask in case this became out of date at some point and it actually is now incorrect.


Supamiu:
This is generated data so editing it won't make a difference :/

mav8557:
That makes sense. Does that mean that this is an error upstream (XIVAPI?) or is this actually correct, and I'm just using this wrong?

Supamiu:
We extract data witrh Kobold for our own data format to be made but this is not a bug indeed.

The way it works ingame is that it takes a range of possible levels for each delivery, meaning that you can get deliveries from another level close to yours.

mav8557:
Gotcha, going to close this then. Thanks!

PR #3232, opened 2026-06-06
feat: add macOS desktop and XIV-on-Mac packet capture support

spidermonky0007:
Adds the remaining macOS desktop support on top of the shared cross-platform bridge work now in staging, including Mac app/DMG build scripts and XIV-on-Mac packet capture support.

## Mac desktop/build support

- Adds mac Electron build scripts for .app/DMG builds:
  - yarn electron:setup:build-mac
  - yarn electron:setup:build-mac:dir
- Adds macOS app icon configuration.
- Updates mac packaging to include required files from deucalion-bridge.
- Adds a helper script to download bcryptprimitives.dll without committing the binary.

## XIV-on-Mac packet capture support

- Uses the XIV-on-Mac ffxivConfig directory for DAT watching when present.
- Handles macOS fs.watch rename events for DAT updates.
- Allows the macOS bridge to start before ffxiv_dx11.exe, so the bridge can wait for the game process.
- Stages deucalion.dll and bcryptprimitives.dll into the XIV-on-Mac Wine prefix under C:\deucalion.
- Uses the XIV-on-Mac Wine environment needed for bridge startup.
- Stops macOS packet capture before installing updates so the bridge process is cleaned up before quitAndInstall.

## Validation

- Rebased onto current ffxiv-teamcraft:staging successfully.
- yarn tsc -p apps/electron/tsconfig.app.json --noEmit passed.
- yarn electron:setup:build-mac completed successfully.
- Packaged macOS app/DMG built successfully.
- Local XIV-on-Mac packet capture test confirmed:
  - Teamcraft starts the bridge before the game.
  - Bridge waits for ffxiv_dx11.exe.
  - deucalion.dll injects.
  - Deucalion pipe opens.
  - Bridge listens on 127.0.0.1:31594.
  - Teamcraft connects through bridgeTcpPort.
  - Inventory updates live.


package.json
```diff
@@ -108,11 +110,14 @@
     "electron:copyfiles": "copyfiles ./libs/loader/tc-loader.js ./dist/apps/electron/ -f",
     "electron:build": "nx run electron:build && yarn electron:copyfiles && tsc -p ./apps/electron/tsconfig.preload.json",
     "electron:start": "yarn electron:build && yarn electron ./dist/apps/electron/main.js",
+    "download:bcryptprimitives": "node ./tools/build/download-bcryptprimitives.js",
```
birkholz: While it's slightly convenient to download the latest version at build time, I think it's generally much safer to lock in a specific trusted version and check in it with git, same as I did for the bridge exe.
That way, changes are known and expected, not unexpected suddenly as a side effect of a new version being built.

apps/electron/src/pcap/packet-capture.ts
```diff
@@ -274,7 +274,7 @@ export class PacketCapture {
         // With some launchers, the bridge process (Wine) prevents the game from
         // booting if it starts before ffxiv_dx11.exe is already running.  Fail
         // fast with a clear error so the user knows to launch the game first.
-        if (!this.isGameRunningViaWine()) {
+        if (process.platform !== 'darwin' && !this.isGameRunningViaWine()) {
```
birkholz, line 277: I feel like we talked about this before, but remind me why this change is here?

apps/electron/src/dat/dat-files-watcher.ts
```diff
@@ -185,6 +185,12 @@ export class DatFilesWatcher {
       // Nothing found
       return null;
     }
+    if (process.platform === 'darwin') {
+      const xivOnMacConfigDir = join(app.getPath('appData'), 'XIV on Mac', 'ffxivConfig');
+      if (existsSync(xivOnMacConfigDir)) {
+        return xivOnMacConfigDir;
+      }
```
birkholz, line 192: Can you also detect the official launcher?

apps/electron/src/update/auto-updater.ts
```diff
@@ -77,8 +77,17 @@ export class AutoUpdater {
...
+    ipcMain.on('install-update', async () => {
       (<any>app).isQuitting = true;
+
+      if (process.platform === 'darwin') {
+        try {
+          await this.pcap.stop();
+        } catch (e) {
+          log.error('[updater] Failed to stop packet capture before installing update:', e);
+        }
+      }
```
birkholz: Was this the fix for some bug around updates? The bridge runs as a sub-process (at least on Linux) so it should stop along with the Teamcraft app stopping. Does Mac keep the bridge running when Teamcraft shuts down?

spidermonky0007: Fair point. This was intended as a defensive cleanup before quitAndInstall, but I do not currently have a confirmed reproduced bug showing that macOS leaves the bridge process running after Teamcraft exits.

To keep this PR focused, I can remove the updater cleanup change for now and revisit it separately if we can reproduce an update/install issue with the bridge still running.

I was more pre-emptive do to how XoM handles packet capture.

package.json
spidermonky0007: Updated this to avoid downloading bcryptprimitives.dll at build time.

Changes:
- removed the download helper script
- removed the build script dependency on yarn download:bcryptprimitives
- committed the tested bcryptprimitives.dll directly beside deucalion-bridge.exe

Validation:
- yarn tsc -p apps/electron/tsconfig.app.json --noEmit passed
- yarn electron:setup:build-mac passed
- installed from the generated DMG
- installed app contains both deucalion-bridge.exe and bcryptprimitives.dll
- packet capture still reaches the bridge waiting flow and works locally

apps/electron/src/dat/dat-files-watcher.ts
spidermonky0007: This PR currently validates XIV-on-Mac support only.
Official launcher support is desired, but I cannot verify it locally because the official launcher does not complete setup on my machine.

apps/electron/src/pcap/packet-capture.ts
spidermonky0007: This is specifically for the XIV-on-Mac flow I was able to get working.

On macOS, the reliable startup order was:
1. Start Teamcraft / enable packet capture.
2. Spawn deucalion-bridge under XIV-on-Mac Wine.
3. Let the bridge wait for ffxiv_dx11.exe.
4. Start XIV-on-Mac.
5. Bridge detects the process, injects, opens the pipe, and Teamcraft connects.

If we keep the existing isGameRunningViaWine() fail-fast behavior on macOS, Teamcraft refuses to start the bridge before the game exists, which prevents that working flow. The condition is scoped so Windows/Linux keep the existing behavior, while macOS can use the bridge’s built-in waiting behavior.

spidermonky0007:
I also updated the bcryptprimitives handling based on the review feedback. It is now committed directly beside the bridge executable instead of being downloaded during the build.

Validation after that change:
- yarn tsc -p apps/electron/tsconfig.app.json --noEmit passed
- yarn electron:setup:build-mac passed
- installed from the generated DMG
- installed app contains both deucalion-bridge.exe and bcryptprimitives.dll
- XIV-on-Mac packet capture still works locally

Official launcher note:
I found the official launcher Wine binary/prefix locally, but I can’t reliably validate official launcher support because my official launcher install hangs on the install/relaunch prompts. I’m not claiming official launcher packet capture support in this PR unless another tester can validate the official launcher flow.

EDKarlsson:
@spidermonky0007 -- The official launcher is broken on macbooks, at least on my silicon mac it is. The official installer process goes into an infinite install/crash loop.

XIV on Mac is the most reliable way of installing FFXIV on Mac.

spidermonky0007:
> @spidermonky0007 -- The official launcher is broken on macbooks, at least on my silicon mac it is. The official installer process goes into an infinite install/crash loop.
> 
> 
> 
> XIV on Mac is the most reliable way of installing FFXIV on Mac.

Considering this is the general consensus of Mac users, is kinda why I'm just leaving it to a volunteer to work with me if they want official support lol, but I'm happy with just XoM for now. It's up to Miu though if they are ok with that to merge or not

EDKarlsson:
For sure. Do you want to close my pr? My solution is a bit different and doesn't need to include a bin/lib dependency.

spidermonky0007:
> For sure. Do you want to close my pr? My solution is a bit different and doesn't need to include a bin/lib dependency.

Message me on discord in the Teamcraft public devs chat, and lets talk about it or maybe even collab/merge the work together if there is value on both ends. I am still new to any dev work so I have relied a lot on understanding the work of the devs there and what not to get any of this work done :)

spidermonky0007:
Going to close this in favor of this PR: https://github.com/ffxiv-teamcraft/ffxiv-teamcraft/pull/3236/

it handles very similarly, and I will continue to help where I can with Official launcher support, as well as helping with any other macos parts as needed.

Thank you @EDKarlsson for helping clean this up!

PR #3236, opened 2026-06-09
feat(desktop): macOS support for the Electron desktop app (XIV on Mac)

EDKarlsson:
## Summary

Adds macOS support to the Electron desktop app, extending the existing Linux/Wine port. On macOS the game runs under Wine via [XIV on Mac](https://www.xivmac.com/) (the equivalent of XIVLauncher on Linux), so the same bridge/file-watching architecture applies — the code just never extended its platform checks or build scripts to `darwin`. Everything below was verified end-to-end on a live XIV on Mac session (Apple Silicon).

## What this enables

- **Builds** into a `.app`/`.dmg` on macOS (previously only Windows `-w` and Linux `--linux` build scripts existed).
- **Runs** — the app launches and the renderer loads.
- **Inventory tracking** — detects the current character and inventory changes.
- **Packet capture** — Deucalion injects through XIV on Mac's Wine and hooks packets.
- **Google login** — works in production builds.

## Screenshot

image

## Changes

**Build**
- Add `build:dmg` / `build:dmg:dir` scripts and a complete `mac` electron-builder config (`dmg` target, icon, artifactName). Added a 1024px icon (`tools/build/icon-mac.png`) since macOS requires ≥512px.
- Make `prod-assets-hash.js` idempotent — the extracts rename threw `ENOENT` on any re-run; the data-file loop was already guarded.

**Runtime (`apps/electron`)**
- `dat-files-watcher`: resolve the FFXIV config dir from XIV on Mac (`~/Library/Application Support/XIV on Mac/ffxivConfig`) on `darwin`; and accept `rename` events, not just `change` — the game saves `ITEMODR.DAT` atomically and macOS FSEvents reports only `rename`, so character detection never fired before.
- `packet-capture`: auto-detect XIV on Mac's bundled Wine binary + prefix, and pass `WINEESYNC=1 WINEMSYNC=1 WINEFSYNC=0` when spawning the bridge so it attaches to the game's existing wineserver (a per-prefix server requires matching esync/msync, or injection silently fails).

**Auth (`apps/client`)**
- The desktop Google OAuth chain subscribed with only a `next` handler, so any failure was swallowed and the login modal hung forever. Add error propagation + step logging, and handle the error in the login/register popups. Documented that Google login needs the production build (`build:dmg`) — beta builds fail with `auth/invalid-credential` because the OAuth client is hardcoded to production.

## Notes / scope

- Targets the standard XIV on Mac install — which [does not support custom install paths](https://www.xivmac.com/frequently-asked-questions), so the fixed defaults are correct rather than a shortcut.
- Packet capture is ToS-adjacent and off by default, same as on Windows/Linux.
- Local builds are unsigned (no Apple Developer ID required); electron-builder skips signing gracefully.

## Testing

1. `yarn build:dmg` and launch the `.app`.
2. Log in with Google → succeeds.
3. Enable packet capture, get in-game → Deucalion hooks packets.
4. Sort/unsort inventory → character detected, inventory tracking starts.


EDKarlsson:
Just realized someone already has a PR for MacOS support! Sorry for the duplicate.

EDKarlsson:
Spidermonky chose to close their PR and have my implementation go forward for review.

Supamiu (dismissed):

EDKarlsson:
Thanks for approving the PR @Supamiu I'm working on the merge conflict. It looks like a wine-resolver was implemented on a different PR that I had not pulled in at the time I submitted my PR. I didn't get any feedback so I'm assuming that I should move my wine resolve for mac implementation into `wine-resolver.ts` class. Will try to work on it this week.

EDKarlsson:
Can I get an update on this PR?

Supamiu:
Is everything ready on your end?

PR #3273, opened 2026-08-01
feat(simulator) Added Auto Solver Feature

PawanaMcKay:
# What this adds

This adds a new "Solver" Action Button in the Simulator UI, that automatically generates a complete, guaranteed-successful crafting rotation for the currently configured recipe and crafter stats, instead of requiring the user to build it by hand.
The Solver runs entirely in a dedicated Web Worker (`solver-worker.ts`) so the search never blocks the UI thread and streams live progress (current depth, best quality found so far) back to a new `SolverPopupComponent` while it runs.

## New files

- `solver-core.ts` - the framework for the search algorithm
- `solver.worker.ts` - Web Worker entry point wiring the algorithm to the active Simulator Bundle
- `solver.service.ts` - Service exposing the solver as an observable stream
- `solver-popup.component.ts` (+ template/styles) - progress UI and result application

## How the algorithm works

The solver uses a beam search with branch-and-bound pruning over the space of possible action sequences:
1. **State expansion**. From each currently tracked partial rotation ("node"), every action that is (a) usable in the current simulated state, (b) has a 100% success rate ("safe mode" - never relies on a random condition roll), and (c) is affordable given remaining CP, is tried as a possible next step
2. **Scoring**. Each resulting node is scored so that: a completed rotation (100% progress) always outranks any incomplete one regardless of quality. Among completed rotations, higher quality wins. Among incomplete rotations, progress and quality are weighted equally, with a penalty applied if durability drops dangerously low before the craft is done (to force timely use of Master's Mend / Immaculate Mend)
3. **Deduplication**. Nodes with near-identical simulated state (bucketed CP/durability/progress/quality, plus an exact active-buff signature) are merged, keeping only the highest-scoring one, so the beam isn't wasted on near-duplicate branches.
4. **Pruning**. Branches that cannot possibly reach 100% progress within the remaining step budget - even under an optimistic best-case estimate - are discarded early (after a short grace period, so early buff-setup steps aren't pruned)
5. **Beam limiting**. After each depth, only the top `beamWidth` scoring nodes are kept for the next iteration, bounding the search space
6. **Termination**. The search stops once a completed rotaion reaching full quality is found, the step limit is reached or the time budget expires. As a safety net, if no completed rotation is found at all (should not normally happen given the pruning above), a simple greedy fallback appends the highest-progress-yielding safe action repeatedly until the craft completes.

Because a completed rotation is never expanded further, the craft is guaranteed to reach 100% progress exactly on the final action of the returned rotation, never before or after.

## Configuration

Two boolean flags, currently defaulted to `false` and not yet exposed in the UI (maybe for a follow up work), gate action pools that aren't universally available:
- `shouldUseCosmicExploration` - allows "Cosmic Exploration" actions (Material Miracle, Stellar Steady Hand), which only apply to specific recipes
- `shouldUseSpecialistCommands` - allows Specialist-only actions (Careful Observation, Heart and Soul, Quick Innovation). Only takes effect, if the crafter stats also have `specialist: true`

## Known limitations

- Each simulated node re-runs the full action sequence from scratch rather than incrementally cloning state. Switching to incremental simulation would allow deeper/wider search in the same time budget.
- Search parameters (`beamWidth`, `maxSteps`, `maxComputeTime`) are currently hardcoded defaults. Exposing them as a UI control would let advanced users trade off search time vs. thoroughness.


Supamiu:
Hi, could you please ping me on Discord (Same name there) to explain to me why and how you created this? I'm really interested ^^

Also, does this work in an electron environment? Workers can be funky there.

PawanaMcKay:
> Hi, could you please ping me on Discord (Same name there) to explain to me why and how you created this? I'm really interested ^^
> 
> Also, does this work in an electron environment? Workers can be funky there.

Tested it on Electron, too (after a bit of struggle) and there it also works like intended

Supamiu (approved):

